PR #27, opened 2022-03-22
Updated EXICodec.jar with fix for WeldingDetectionRes message

shalinnijel2:
EXICodec.jar updated with fix for WeldingDetectionRes message. The bug was that the ResponseCode field in the JAXB class was pointing at the wrong definition in AppProtocol schema instead of the definition in MsgDataTypes.


slachiewicz:
where we can find the source repo for EXICodec.jar ?

cienporcien:
Likewise, I'd like to add the -20 ACDP messages to the .jar

PR #50, opened 2022-05-24
fixed some technical terms in the readme

tropxy:


README.md
```diff
@@ -17,8 +17,8 @@ There are two recommended ways of running the project:
 
 ### Running with Docker
 Using Docker has the advantage of starting everything up automatically,
-including certificate generation, tests and linting, as well as creating
-both the SECC and EVCC.
+including certificate generation, tests and linting, as well as spawning
```
danielgordon-switch-ev, line 20: Either is fine in this context, I think -- @nikkigithub1 was there a specific reason to change "spawning" to "creating" in the first place?

README.md
```diff
@@ -151,21 +151,22 @@ The SECC and EVCC have been tested together under:
 ## IPv6 WARNING
 
 For the system to work locally, the network interface needs to have
-an IPv6 link-local address assigned.
+an IPv6 local-link address assigned.
```
danielgordon-switch-ev, line 154: A quick Google search suggested that "link-local" is the correct one -- @tropxy are you sure about this?  According to Wikipedia, "In computer networking, a link-local address is a network address that is valid only for communications within the network segment or the broadcast domain that the host is connected to."

danielgordon-switch-ev (commented):
I was wondering about a couple of these myself.

README.md
tropxy: you are right, it is link-local

README.md
```diff
@@ -192,21 +193,21 @@ The following table provides a few of the available variables:
 | MESSAGE_LOG_EXI   | `False`                       | Whether or not to log the EXI Bytestream messages (only works if log level is set to DEBUG)
 
 
-## Licence
+## License
```
danielgordon-switch-ev, line 196: I think this is fine -- American English is often the standard in a technical context.

danielgordon-switch-ev (approved):

PR #51, opened 2022-05-25
Fix running in docker

ruudvd:
I tested the ISO15118 docker-compose file and ran into a few issues:

1.  ModuleNotFoundError: No module named 'aiofile'
```
secc_1   | Traceback (most recent call last):
secc_1   |   File "/venv/bin/iso15118", line 5, in <module>
secc_1   |     from iso15118.secc.main import run
secc_1   |   File "/venv/lib/python3.10/site-packages/iso15118/secc/main.py", line 5, in <module>
secc_1   |     from iso15118.secc.controller.simulator import SimEVSEController
secc_1   |   File "/venv/lib/python3.10/site-packages/iso15118/secc/controller/simulator.py", line 10, in <module>
secc_1   |     from aiofile import async_open
secc_1   | ModuleNotFoundError: No module named 'aiofile'
```
I fixed this by moving `aiofile` dependency 

2. Cannot assign requested address
```
DEBUG    2022-05-25 11:14:09,547 - asyncio (59): Using selector: EpollSelector
INFO    2022-05-25 11:14:10,257 - iso15118.shared.exi_codec (168): EXI Codec version: 1.55
INFO    2022-05-25 11:14:10,257 - iso15118.secc.comm_session_handler (196): Communication session handler started
INFO    2022-05-25 11:14:10,260 - iso15118.secc.transport.udp_server (110): UDP server socket ready
INFO    2022-05-25 11:14:10,260 - iso15118.secc.transport.udp_server (97): UDP server started at address FF02::1%eth0 and port 15118
ERROR    2022-05-25 11:14:10,260 - iso15118.shared.utils (78): [Errno 99] Cannot assign requested address
Traceback (most recent call last):
  File "/venv/lib/python3.10/site-packages/iso15118/shared/utils.py", line 76, in wait_till_finished
    task.result()
  File "/venv/lib/python3.10/site-packages/iso15118/secc/transport/tcp_server.py", line 39, in start_tls
    await self.server_factory(tls=True)
  File "/venv/lib/python3.10/site-packages/iso15118/secc/transport/tcp_server.py", line 97, in server_factory
    sock.bind(self.full_ipv6_address)
OSError: [Errno 99] Cannot assign requested address
ERROR    2022-05-25 11:14:10,261 - iso15118.shared.utils (78): [Errno 99] Cannot assign requested address
Traceback (most recent call last):
  File "/venv/lib/python3.10/site-packages/iso15118/shared/utils.py", line 76, in wait_till_finished
    task.result()
  File "/venv/lib/python3.10/site-packages/iso15118/secc/transport/tcp_server.py", line 45, in start_no_tls
    await self.server_factory(tls=False)
  File "/venv/lib/python3.10/site-packages/iso15118/secc/transport/tcp_server.py", line 97, in server_factory
    sock.bind(self.full_ipv6_address)
OSError: [Errno 99] Cannot assign requested address
ERROR    2022-05-25 11:14:10,261 - iso15118.secc (35): SECC terminated: 
```
After i added a sleep of 1 sec before running the executable i got this:
the `TLS server started at ..` and `TCP server started at ..` is the docker container network not yet started when the execuable is ran?

### used setup
Pop!_OS 21.10 
Docker version 20.10.12, build e91ed57
docker-compose version 1.29.2, build 5becea4c

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ cryptography = "3.4.6"
 pydantic = "^1.9.0"
 psutil = "^5.9.0"
 py4j = "^0.10.9"
-
+aiofile = "^3.7.4"
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1.1"
@@ -30,7 +30,6 @@ black = "22.3.0"
 isort = "^5.10.1"
 flake8 = "^4.0.1"
 pytest-cov = "^3.0.0"
-aiofile = "^3.7.4"
 
 [tool.isort]
 profile = "black"
--- a/template.Dockerfile
+++ b/template.Dockerfile
@@ -71,4 +71,4 @@ COPY --from=build /usr/src/app/iso15118/shared/pki/ /usr/src/app/iso15118/shared
 
 
 # This will run the entrypoint script defined in the pyproject.toml
-CMD /venv/bin/iso15118
+CMD sleep 1 && /venv/bin/iso15118


ruudvd:
> Hi ruudvd, thanks for trying out our project and for the feedback.
> 
> Sorry about the error with aiofile package. It happens that this package was added to the pyproject.toml as a dev dependency and in fact is in the right place as aiofile is used in the simulator.py, which is just an example and must not be treated as part of the package.
> 
> I opened a PR where I install the aiofile on running time, avoiding adding the package as a prod dependency.
> 
> Regarding the Docker error, I must say I also experienced the same Error in MacOS, something I didnt experience before. But in my case, anyway, I could run the container after the 1st build, however in the upcoming runs, it would fail to bind to the IPv6 address. What I found out is that between container runs, Docker always assigns the same IPv6 for the same interface, so my suspicion is that Docker network does not "forget" about the binding with the previous container run and when we try to run for multiple times, Docker says that the address is already used.
> 
> As a workaround what I do is after stopping the container, to rm the stopped containers with `docker rm $(docker ps -qa)` and then `make dev`. No need to rebuild the containers as the image is still in cache. For now that is what I can tell you, I hope it helps.

I tested the PR with the changed aiofile and the docker workaround, it worked. 
The EVCC keeps getting TimeoutErrors till the SECC downloaded and installed aiofile.

tropxy:
> > Hi ruudvd, thanks for trying out our project and for the feedback.
> > Sorry about the error with aiofile package. It happens that this package was added to the pyproject.toml as a dev dependency and in fact is in the right place as aiofile is used in the simulator.py, which is just an example and must not be treated as part of the package.
> > I opened a PR where I install the aiofile on running time, avoiding adding the package as a prod dependency.
> > Regarding the Docker error, I must say I also experienced the same Error in MacOS, something I didnt experience before. But in my case, anyway, I could run the container after the 1st build, however in the upcoming runs, it would fail to bind to the IPv6 address. What I found out is that between container runs, Docker always assigns the same IPv6 for the same interface, so my suspicion is that Docker network does not "forget" about the binding with the previous container run and when we try to run for multiple times, Docker says that the address is already used.
> > As a workaround what I do is after stopping the container, to rm the stopped containers with `docker rm $(docker ps -qa)` and then `make dev`. No need to rebuild the containers as the image is still in cache. For now that is what I can tell you, I hope it helps.
> 
> I tested the PR with the changed aiofile and the docker workaround, it worked. The EVCC keeps getting TimeoutErrors till the SECC downloaded and installed aiofile.

yup, there are few number of times that we can send an SDP request until the system actually crashes, so what you see is the EVCC starting first and has to wait for the SECC to be receptive to process the SDP request.

We have set in docker compose a requirement for the EVCC to start after the SECC does, but that is just related to the container startup, not to the application startup.

PR #55, opened 2022-05-27
added: import of aiofiles and package installation on run time 

tropxy:
as aiofiles is a dev dependency

### DIFF
--- a/iso15118/secc/controller/simulator.py
+++ b/iso15118/secc/controller/simulator.py
@@ -2,12 +2,22 @@
 This module contains the code to retrieve (hardware-related) data from the EVSE
 (Electric Vehicle Supply Equipment).
 """
+
 import logging
 import time
 from dataclasses import dataclass
 from typing import List, Optional
 
-from aiofile import async_open
+
+from iso15118.shared.utils import install
+
+try:
+    from aiofile import async_open
+except ModuleNotFoundError:
+    install("aiofile")
+    from aiofile import async_open
+
+
 from pydantic import BaseModel, Field
 
 from iso15118.secc.controller.interface import EVSEControllerInterface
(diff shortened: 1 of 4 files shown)


shalinnijel2:
Just one minor change that could help with an edge case...else lgtm.

iso15118/secc/controller/simulator.py
```diff
@@ -2,12 +2,22 @@
...
 from typing import List, Optional
 
-from aiofile import async_open
+
+from iso15118.shared.utils import install
+
+try:
+    from aiofile import async_open
+except ModuleNotFoundError:
+    install("aiofile")
```
danielgordon-switch-ev, line 17: Is there a reason not to just move this out of `dev-dependencies` and into `dependencies` in the pyproject.toml file?  We really shouldn't be installing things in one-off ways like this, if we can avoid it at all.

HugoJP1: Poetry 1.2 is now in beta (https://pypi.org/project/poetry/#history)
image

pyproject.toml
```diff
@@ -18,7 +18,8 @@ cryptography = "3.4.6"
 pydantic = "^1.9.0"
 psutil = "^5.9.0"
 py4j = "^0.10.9"
-
+black = "^22.3.0"
+autopep8 = "^1.6.0"
```
HugoJP1, line 22: Shouldn't these two be in the dev dependencies?

iso15118/secc/controller/simulator.py
HugoJP1: In the current version of Poetry there is another way of doing this which is by using [extras](https://python-poetry.org/docs/pyproject#extras). So we would put it in the dependencies, but add it as optional for now. Will have a play around and see if this is a solution which could work.

tropxy:
I just added aiofile installation to the dockerfile in #130

PR #59, opened 2022-06-08
fix: Sum of durations in schedule entries must match departure_time from EVCC (AB#2183)

shalinnijel2:
- Added a validation function to check if the schedule returned by the secondary actor is valid
- V2G2-303/304 doesn't seem to be strict though - as V2G2-305 provides provision to EVCC  to renegotiate a new schedule in case duration/sales tariff needs are not met.
- Added tests for V2G2-303, V2G2-304 and V2G2-761

### DIFF
--- a/iso15118/secc/controller/simulator.py
+++ b/iso15118/secc/controller/simulator.py
@@ -3,6 +3,7 @@ This module contains the code to retrieve (hardware-related) data from the EVSE
 (Electric Vehicle Supply Equipment).
 """
 import logging
+import math
 import time
 from dataclasses import dataclass
 from typing import List, Optional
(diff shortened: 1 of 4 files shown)


tests/secc/states/test_iso15118_2_states.py
```diff
@@ -77,3 +79,110 @@ async def test_authorization_next_state_on_authorization_request(
...
+            if schedule_tuples.p_max_schedule.schedule_entries is not None:
+                start_time = schedule_tuples.p_max_schedule.schedule_entries[
+                    0
+                ].time_interval.start
+
+            for entry in schedule_tuples.p_max_schedule.schedule_entries:
+                schedule_duration += entry.time_interval.start - start_time
+                if entry.time_interval.duration is not None:
+                    schedule_duration += entry.time_interval.duration
+                start_time = entry.time_interval.start
```
tropxy, line 122: following the same comment as before, I think we can get the last and first entry and do the calculation

tests/secc/states/test_iso15118_2_states.py
```diff
@@ -77,3 +79,110 @@ async def test_authorization_next_state_on_authorization_request(
...
+        sa_schedule_tuples = (
+            charge_parameter_discovery_res.sa_schedule_list.schedule_tuples
+        )
+
+        for schedule_tuples in sa_schedule_tuples:
+            assert schedule_tuples.p_max_schedule is not None
+            found_entry_indicating_start_without_delay = False
+            duration_indicating_immediate_start = 30
+            for entry in schedule_tuples.p_max_schedule.schedule_entries:
+                if entry.time_interval.start < duration_indicating_immediate_start:
```
tropxy, line 184: where does this 30 secs value comes from? is it a recommendation?
I think the start entry should have a start value = 0

tests/secc/states/test_messages.py
```diff
@@ -115,3 +123,53 @@ def get_dummy_v2g_message_authorization_req():
...
+
+    return V2GMessage(
+        header=MessageHeader(session_id=MOCK_SESSION_ID),
+        body=Body(charge_parameter_discovery_req=charge_parameter_discovery_req),
+    )
+
+
+def get_charge_parameter_discovery_req_message_no_departure_time():
+    e_amount = PVEAmount(multiplier=0, value=60, unit=UnitSymbol.WATT_HOURS)
+    ev_max_voltage = PVEVMaxVoltage(multiplier=0, value=400, unit=UnitSymbol.VOLTAGE)
```
tropxy, line 156: can you add these to a conftest.py file and add the following decorator on each method: "@pytest.fixture"

you can then use directly the return of that method into the test_ method you use. For example:

```python
@pytest.fixture
def get_charge_parameter_discovery_req_message_no_departure_time():
    ....
```

```python
async def test_charge_parameter_discovery_res_v2g2_761(self, discover_req_message_no_dep_time):
    # V2G2-761: If departure time was not provided, then SECC shall assume
    # that the EV intends to start charging without any delay
    charge_parameter_discovery = ChargeParameterDiscovery(self.comm_session)
    charge_parameter_discovery.process_message(
        message=discover_req_message_no_dep_time
    )
    assert (
        charge_parameter_discovery.next_msg.body.charge_parameter_discovery_res
        is not None
    )
```

tropxy (changes requested):
check it out

iso15118/secc/controller/simulator.py
```diff
@@ -204,9 +205,9 @@ def get_supported_energy_transfer_modes(
             return [dc_extended]
 
         # ac_single_phase = EnergyTransferModeEnum.AC_SINGLE_PHASE_CORE
-        # ac_three_phase = EnergyTransferModeEnum.AC_THREE_PHASE_CORE
+        ac_three_phase = EnergyTransferModeEnum.AC_THREE_PHASE_CORE
         dc_extended = EnergyTransferModeEnum.DC_EXTENDED
-        return [dc_extended]
+        return [dc_extended, ac_three_phase]
```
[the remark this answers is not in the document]
shalinnijel2: Done.

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1062,6 +1076,56 @@ def process_message(
...
+                logger.warning(f"Schedule tuples {schedule_tuples}")
+                valid = False
+                break
+
+            # Not setting this check as equality check as it is possible that the time
+            # could be off by few seconds. Also considering V2G2-305, it would suffice
+            # if departure_time_total is at least the same as departure_time
+            elif departure_time != 0 and departure_time < schedule_duration:
+                valid = False
+                break
```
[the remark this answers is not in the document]
shalinnijel2: Sure, done.

tests/secc/states/test_iso15118_2_states.py
shalinnijel2: Same as above - have done the mod.

tests/secc/states/test_iso15118_2_states.py
shalinnijel2: I was doubtful about this - added 30 as a rough indicator of immediate start - have set it to 0.

tropxy (approved):

PR #60, opened 2022-06-09
Make communication protocols configurable via .env file

shalinnijel2:
Modified secc_settings.py to read protocols and auth_modes from .env file if set.
If not set, then would default to all supported protocols.
This mod only applies to secc_settings - evcc protocol preferences are still hardcoded in evcc_settings.py
This would help test other use cases like evcc having different protocol priorities to secc.
The env variable is added and left commented out in the .env.dev.local and .env.dev.docker file to hint that configuring protocols is possible.


iso15118/secc/secc_settings.py
```diff
@@ -70,22 +87,43 @@ def load_envs(self, env_path: Optional[str] = None) -> None:
...
 
         # Whether the charging station allows the EV to go into Standby (one of the
         # enum values in PowerDeliveryReq's ChargeProgress field). In Standby, the
         # EV can still use value-added services while not consuming any power.
         self.standby_allowed = env.bool("STANDBY_ALLOWED", default=False)
 
         env.seal()  # raise all errors at once, if any
+
+    def load_requested_protocols(self, read_protocols: Optional[List[str]]):
+        protocols = format_list(read_protocols)
```
tropxy, line 108: is there any exception we shall catch? or the failure is reason to break the code and dont go further?

iso15118/secc/secc_settings.py
```diff
@@ -70,22 +87,43 @@ def load_envs(self, env_path: Optional[str] = None) -> None:
...
+        self.supported_protocols = []
+        for protocol in protocols:
+            if protocol in self.default_protocols:
+                self.supported_protocols.append(Protocol[protocol])
+            else:
+                logger.debug(
+                    f"Unknown protocol {protocol}. Supported protocols:"
+                    f" {self.default_protocols}"
+                )
```
tropxy, line 117: you can do the following:
```python
self.supported_protocols = list(set(protocols).intersection(self.default_protocols)
if self.supported_protocols is None:
    raise ...
```

I think we shall raise an exception in case there are no protocols requested that the system actually supports

iso15118/secc/secc_settings.py
```diff
@@ -70,22 +87,43 @@ def load_envs(self, env_path: Optional[str] = None) -> None:
...
+    def load_requested_auth_modes(self, read_auth_modes: Optional[List[str]]):
+        auth_modes = format_list(read_auth_modes)
+        self.supported_auth_options = []
+        for auth_mode in auth_modes:
+            if auth_mode in self.default_auth_modes:
+                self.supported_auth_options.append(AuthEnum[auth_mode])
+            else:
+                logger.debug(
+                    f"Unknown authentication mode {auth_mode}. Supported"
+                    f" authentication modes: {self.default_auth_modes}"
```
tropxy, line 128: the comments I did above also apply here

tropxy (changes requested):

iso15118/secc/secc_settings.py
shalinnijel2: Sure, the possibility you have listed below is one.

iso15118/secc/secc_settings.py
shalinnijel2: Sure, done.

iso15118/secc/secc_settings.py
shalinnijel2: Sure, done.

iso15118/secc/secc_settings.py
shalinnijel2: I would have preferred to initialise default_protocols as `default_protocols = Protocol.names()` but had to avoid that as not all the ones that would be listed are currently supported (for instance 15118_20_WPT).

shalinnijel2:
> Dont forget to modify the README as well with the new ENV

Sure, done.

tropxy:
rebase the branch and I think is good to go

tropxy (approved):

PR #65, opened 2022-06-20
fixed order of closing contactor and reformated the code

tropxy:


ikaratass (commented):
it is ok for me

iso15118/secc/states/din_spec_states.py
```diff
@@ -644,14 +644,19 @@ async def process_message(
                 "Stay in this state and expect "
                 "WeldingDetectionReq/SessionStopReq"
             )
+            await self.comm_session.evse_controller.set_hlc_charging(False)
             next_state = None
-            await self.comm_session.evse_controller.open_contactor()
+
```
shalinnijel2, line 649: I understand you were looking at the closing contactor side of things, but I see contactor is not closed in din_spec. Do you want to paste the below lines under the `if power_delivery_req.ready_to_charge:` block:
```
            contactor_state = await self.comm_session.evse_controller.close_contactor()
            if contactor_state != Contactor.CLOSED:
                self.stop_state_machine(
                    "Contactor didn't close",
                    message,
                    ResponseCode.FAILED,
                )
                return
```

shalinnijel2 (changes requested):
Just that close_contactor is missing in din_spec.
So, to summarise, the order of events for all protocols would be:
1. Start/Stop HLC
2. Stop charger (only if ChargeProgress equals stop charging)
3. Open/Close contactor

iso15118/secc/states/din_spec_states.py
tropxy: It actually has to happen during CableCheck in the DC scenario... I will add it there, for both -2 and Din

PR #69, opened 2022-06-23
CharINTestivalPrep 2022: Private key password modified

shalinnijel2:
SHOULD NOT MERGE TO MASTER.
Private key decode password modified as per requirement for the CharIN testival in Poland 2022.
Temporary workaround for the testival. WIll soon add a task to handle this better.


tropxy:
Closed as release #72 contains a change that enables to read the password from an env file

PR #73, opened 2022-06-28
added id to the Reference structure; notes on PnC and fix signature verification

tropxy:
as even so it is optional must be…


iso15118/secc/secc_settings.py
```diff
@@ -38,6 +38,7 @@ class Config:
         "DIN_SPEC_70121",
         "ISO_15118_2",
         "ISO_15118_20_AC",
+        "ISO_15118_20_DC",
```
shalinnijel2, line 41: This could be removed as it is not fully supported.

iso15118/shared/examples/15118_20_evse_service_config.json
```diff
@@ -177,6 +177,39 @@
...
+            {
+              "Name": "GeneratorMode",
+              "intValue": 1
+            }
+          ]
+        }
+      ]
     }
   }
-}
```
shalinnijel2, line 182: We could perhaps remove this as -20 DC is not fully supported.

shalinnijel2 (changes requested):

iso15118/evcc/controller/simulator.py
```diff
@@ -452,7 +452,7 @@ async def process_dynamic_se_params(
 
     async def is_cert_install_needed(self) -> bool:
         """Overrides EVControllerInterface.is_cert_install_needed()."""
-        return False
+        return True
```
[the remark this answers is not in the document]
tropxy: that is indeed the current boolean that we return

iso15118/secc/secc_settings.py
tropxy: what if we keep it and add a note that DC is under development yet?

iso15118/shared/examples/15118_20_evse_service_config.json
shalinnijel2: Same comment as above - If this is left it in, then the missing parameter set for control mode 2 in DC BPT could be added back in ...

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -586,6 +586,11 @@ async def process_message(
             msg.body.certificate_installation_req
         )
 
+        # For the CertificateInstallation, the min the EVSE can do is
```
MarcMueltin, line 589: Let's replace "EVSE" with "SECC" to be a bit more precise. (As the EVSE is just the power delivery unit, not the communication controller)

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -586,6 +586,11 @@ async def process_message(
...
+        # to verify the message signature, using the OEM provisioning
+        # certificate (public key)
+        # The chain of certificates of the OEM Root can also be verified,
```
MarcMueltin, line 592: Slight clarification: you don't verify certificates, you verify signatures. I'd write:
"The chain of signatures, from the signature in the CertificateInstallationReq's header all the way to the self-signed OEM/V2G root certificate, can be verified if the OEM Sub-CA and OEM/V2G root CA certificate is available."

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -890,6 +911,11 @@ async def process_message(
                 )
                 return
 
+        # TODO: is_authorized needs, in case of PnC, needs to accept as arguments
```
MarcMueltin, line 914: Rewrite to: "In case of PnC, is_authorized needs to accept ..."

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -890,6 +911,11 @@ async def process_message(
...
+        # the EMAID, the contract certificate chain (leaf, MO-Sub-1 and MO-Sub-2,
+        # and the OCSP iso15118CertificateHashData as mentioned in OCPP 2.0.1
+        # use case C07 the contract chain is saved within
```
MarcMueltin, line 917: Start new sentence with: "The contract certificate chain is saved ..."

iso15118/shared/security.py
```diff
@@ -813,22 +833,46 @@ def verify_signature(
     # 2. Step: Checking signature value
     logger.debug("Verifying signature value for SignedInfo element")
     pub_key = load_der_x509_certificate(leaf_cert).public_key()
+
+    # The signature value element corresponds to the encryption of the EXI encoded
```
MarcMueltin, line 837: That needs to be phrased more clearly:
"The signature value element corresponds to the encryption of the EXI encoded and then hashed signed_info element. 
Signed Info -> EXI encoding -> Hashing -> Encryption with private key => Signature Value
ATTENTION: The hashing and encryption operation is part of the ECDSA (Elliptic Curve Digital Signature Algorithm) operation. That is why we do NOT additionally hash the EXI encoded signed info element."

iso15118/shared/security.py
```diff
@@ -813,22 +833,46 @@ def verify_signature(
...
+    # signed_info element.
+    # Signed Info -> EXI encoding -> Encryption with Private Key => Signature Value
     exi_encoded_signed_info = EXI().to_exi(signature.signed_info, Namespace.XML_DSIG)
 
+    # The verify method from cryptography expects the signature to be in DER encoded
+    # format. Please, check: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey.verify  # noqa: E501
```
MarcMueltin, line 843: Get rid of the comma after "Please, "

iso15118/shared/security.py
```diff
@@ -813,22 +833,46 @@ def verify_signature(
...
+    # However, in ISO15118 the signature value is exchanged in raw format.
```
MarcMueltin, line 844: Can we please write "ISO 15118" with a space?

iso15118/shared/security.py
tropxy: done

iso15118/secc/secc_settings.py
tropxy: pls check ;)

iso15118/shared/security.py
tropxy: done

shalinnijel2 (approved):
Okie dokie...approved...

MarcMueltin (approved):

PR #79, opened 2022-07-13
docs: makefile comments and cleanup

mdwcrft:
Adds comments to makefile commands which can be picked up by external tools like [fig](https://fig.io):

Before:
Screenshot 2022-07-13 at 09 47 02

After:
Screenshot 2022-07-13 at 09 47 25

Also removed the unused `docs`, and `black` in `code-quality` which isn't needed as it's part of `reformat`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ export PATH := ${HOME}/.local/bin:$(PATH)
 
 IS_POETRY := $(shell pip freeze | grep "poetry==")
 
-
+# Output descriptions of all commands
 help:
 	@echo "Please use 'make <target>', where <target> is one of"
 	@echo ""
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -58,42 +58,55 @@ build: .generate_v2_certs
...
 black:
 	black --check --diff --line-length=88 iso15118 tests
 
+# Run flake8 checks
 flake8:
 	flake8 --config .flake8 iso15118 tests
 
-code-quality: reformat mypy black flake8
+# Run black, isort, mypy, & flake8
+code-quality: reformat mypy flake8
```
shalinnijel2, line 107: Was black accidentally removed?

mdwcrft: `reformat` runs black so we were actually running it twice here

shalinnijel2 (approved):
All lgtm.

PR #80, opened 2022-07-14
AB#2440 : Add option to fetch CertificateInstallationRes from backend

shalinnijel2:
A summary of the changes:

- A new env variable USE_CPO_BACKEND has been added. Default value is set to False for now as users are likely to run 15118 repo locally. I think in josev_pro we can have this default set to True
- Signature of process_message has been updated to pass through the exi representation of the message. It is set as optional at the moment though the exi stream is passed in to all states except from the tests. All implementations updated.
- A new method get_15118_ev_certificate has been added to SECC interface. The implementation in the SECC simulator does the same as what we do when certificates are available locally - only difference being that the response CertificatesInstallationRes message is base64 encoded version of it's EXI to emulate backend behaviour
- Also, the priority of the protocols in EVCC has been updated - 15118-2 takes priority over 15118-20 as users of the repo are more likely to be interested in 15118-2 implementation at the moment.


iso15118/secc/controller/simulator.py
```diff
@@ -751,3 +774,134 @@ async def get_dc_bpt_charge_params_v20(
             evse_max_discharge_current=RationalNumber(exponent=0, value=11),
             evse_min_discharge_current=RationalNumber(exponent=0, value=0),
         )
+
+    async def get_15118_ev_certificate(
+        self, base64_encoded_cert_installation_req: str, namespace: str
+    ) -> (str, bool, str):
+        """Overrides EVSEControllerInterface.get_15118_ev_certificate()."""
+        """
```
mdwcrft, line 782:
```suggestion
        """Overrides EVSEControllerInterface.get_15118_ev_certificate().
```

.env.dev.docker
```diff
@@ -9,6 +9,7 @@ REDIS_HOST=redis
 FREE_CHARGING_SERVICE=False
 FREE_CERT_INSTALL_SERVICE=True
 ALLOW_CERT_INSTALL_SERVICE=True
+USE_CPO_CERT_INSTALL_SERVICE=False
```
mdwcrft, line 12: Could this be more generic like `CPO_BACKEND_AVAILABLE`? Then we can use it to determine whether to mock the CSMS responses for all scenarios like this

iso15118/secc/controller/simulator.py
```diff
@@ -751,3 +774,131 @@ async def get_dc_bpt_charge_params_v20(
...
+        )
+
+        # base64.b64encode in Python is a binary transform so the return value is byte[]
+        # But the CPO expects exi_encoded_cert_installation_res as a string, hence the
+        # added .decode("utf-8")
+        base64_encode_cert_install_res = base64.b64encode(
+            exi_encoded_cert_installation_res
+        ).decode("utf-8")
+
+        return base64_encode_cert_install_res
```
tropxy, line 904: super cool, @shalinnijel2 ! 😎

iso15118/shared/states.py
```diff
@@ -264,6 +269,7 @@ def create_next_message(
             except ValidationError as exc:
                 logger.exception(exc)
                 raise exc
+            self.next_msg = to_be_exi_encoded
```
tropxy, line 272: This `next_msg` attribute of the State class is a bit useless and I think has even the wrong name as the purpose was to save the last message sent, I believe...for purposes of debugging, but we are not using it for that purpose. So, at least I would change the name...what do you think?

Also this `self.next_msg` and `next_msg` just makes things confusing...

tropxy (changes requested):

iso15118/shared/states.py
shalinnijel2: Yes, I felt the same. Initially I thought we could remove next_msg completely but then realised it is being used in tests (and will be of use in the future) and also in comm_session as follows just before V2GTP packet is sent:
```
         Args:
            message: A V2GTPMessage
        """
        logger.info(f"Sending {str(self.current_state.next_msg)}")
```
I felt it a bit odd to read the above as "sending next message from current state", rather than reading "sending message from current state" as this is the message output from "current_state"

I've now renamed the field to "message". Do you think that reads a bit better?

Also came across a bug where the above `logger.info` was printing `Sending None` with the new changes.
So have added a Base64 "wrapper" to hold the base64 CertificateInstallationRes to help with this.

iso15118/shared/states.py
```diff
@@ -290,24 +311,34 @@ def create_next_message(
...
+            self.message = to_be_exi_encoded
 
-        self.next_msg = to_be_exi_encoded
-
-        # If either next_msg or next_msg_payload_type are None, the state's
+        # If either message or next_msg_payload_type are None, the state's
         # attribute next_v2gtp_msg will not be set. This causes the state
         # machine to raise a FaultyStateImplementationError if next state is
         # not set to Terminate, so no need to raise anything here.
         if next_msg and next_msg_payload_type:
```
tropxy, line 328: we actually dont need this next_msg check...we can use just the `to_be_exi_encoded`.
and we do:

```python
elif isinstance(next_msg, Base64):
            # Incoming message is base64 encoded EXI message.
            # So a base64 decode should retrieve the EXI stream.
            self.message = next_msg
            exi_payload = base64.b64decode(next_msg.message)
else:
    to_be_exi_encoded = next_msg
    self.message = to_be_exi_encoded

    # If either message or next_msg_payload_type are None, the state's
    # attribute next_v2gtp_msg will not be set. This causes the state
    # machine to raise a FaultyStateImplementationError if next state is
    # not set to Terminate, so no need to raise anything here.
    if to_be_exi_encoded and next_msg_payload_type:
          try:
              exi_payload = EXI().to_exi(to_be_exi_encoded, namespace)
          except EXIEncodingError as exc:
              logger.error(f"{exc}")
              self.next_state = Terminate
              raise
```

tropxy: I also would like then to rename the `next_msg` to `msg`...but maybe that is too much of a work for now...

tropxy (changes requested):

iso15118/shared/states.py
shalinnijel2: Done.

iso15118/shared/states.py
```diff
@@ -290,26 +312,33 @@ def create_next_message(
...
                 logger.error(f"{exc}")
                 self.next_state = Terminate
                 raise
 
-            # Step 4
+        # If exi_payload is None, next_v2gtp_msg will not be set.
+        # This in turn, will raise FaultyStateImplementationError if next state is
+        # not set to Terminate, so no need to raise anything here.
+        # Step 4
+        if exi_payload:
```
tropxy, line 341: if there is a `to_be_exi_encoded` then there will be a `exi_payload`

The only case where there is no `to_be_exi_encoded` is when we have a base64 message. but in that case we have a `exi_payload`.
This covers all the cases, so I dont think we need this check

tropxy (changes requested):

iso15118/shared/states.py
shalinnijel2: Hmm...actually in the case of base64 response `to_be_exi_encoded` is not set (as it is not to be exi_encoded) - in this case exi_payload is set directly by base64 decoding the response. This is why the additional `if exi_payload:` was added...

tropxy: correct, `to_be_exi_encoded` is not set, but `exi_payload` will be: https://github.com/SwitchEV/iso15118/pull/80/files#diff-021c1e2405f51ff9a35df9843901c0ffbfe3ea77a23a2bf2844826f7cf695335R320

so, the code in here: https://github.com/SwitchEV/iso15118/pull/80/files#diff-021c1e2405f51ff9a35df9843901c0ffbfe3ea77a23a2bf2844826f7cf695335R346
can run, isnt that right?

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -577,61 +592,120 @@ async def process_message(
...
+                )
+                certificate_installation_res: Base64 = Base64(
+                    payload=base64_certificate_installation_res,
+                    payload_type=CertificateInstallationRes.__name__,
+                )
+            else:
+                (
+                    certificate_installation_res,
+                    signature,
+                ) = self.generate_certificate_installation_res()
```
tropxy, line 644: shouldnt we also check that the all the certificates needed for the verification are present? and in case they are not we raise an exception

iso15118/shared/states.py
shalinnijel2: Ah kk..correct... :) ...have removed the check.

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -577,61 +592,120 @@ async def process_message(
...
+        except Exception as e:
```
tropxy, line 645: we should be strict about the exceptions that can be raised. I guess the ones it can be raised are: 
Timeout, a base64encoding exception, EncryptionError, PrivateKeyReadError and a file not found Error

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -577,61 +592,120 @@ async def process_message(
...
+                # The response received below is EXI response in base64 encoded form.
+                # Decoding to EXI happens later just before V2GTP packet is built.
+                base64_certificate_installation_res = (
+                    await self.comm_session.evse_controller.get_15118_ev_certificate(
+                        base64_certificate_install_req, Namespace.ISO_V2_MSG_DEF
+                    )
+                )
+                certificate_installation_res: Base64 = Base64(
+                    payload=base64_certificate_installation_res,
+                    payload_type=CertificateInstallationRes.__name__,
```
tropxy, line 638: is this what we want? in fact why do we need that `payload_type` field? is not used anywhere and I was confusing it with the `next_msg_payload_type` from the `create_next_message` method

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
shalinnijel2: Yes, Currently the exceptions that could be raised by generate_certificate_installation_res() are EncryptionError, PrivateKeyReadError and FileNotFound. The specifically caught in generate_certificate_installation_res() and raised to be caught here. Is there benefit to catching them specifically here once again?

iso15118/secc/states/iso15118_2_states.py
shalinnijel2: Same as above - currently the exceptions that could be raised by generate_certificate_installation_res() are EncryptionError, PrivateKeyReadError and FileNotFound. The specifically caught in generate_certificate_installation_res() and raised to be caught here. Is there benefit to catching them specifically here once again?

iso15118/secc/states/iso15118_2_states.py
shalinnijel2: This was added to help with logging in the following line. At the moment the only purpose of that field is to help understand what type of message this is when cast to str as is the behaviour with the other message types.

```
    def __str__(self):
        return self.message_type
``` 


https://github.com/SwitchEV/iso15118/blob/4e141ab2c9970f7d8286972cc79354b880303952/iso15118/shared/comm_session.py#L396

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -577,61 +592,120 @@ async def process_message(
...
             return
 
-        cert_install_req: CertificateInstallationReq = (
-            msg.body.certificate_installation_req
+        if not self.validate_message_signature(msg):
+            self.stop_state_machine(
+                "Signature verification failed for " "CertificateInstallationReq",
+                message,
+                ResponseCode.FAILED_SIGNATURE_ERROR,
+            )
```
tropxy, line 606: we need to return here

iso15118/secc/states/iso15118_2_states.py
tropxy: maybe we can `s/payload_type/message_name`

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -577,61 +592,120 @@ async def process_message(
...
+        if (
+            os.path.exists(CertPath.OEM_SUB_CA2_DER)
+            and os.path.exists(CertPath.OEM_SUB_CA1_DER)
+            and os.path.exists(CertPath.OEM_ROOT_DER)
+        ):
+            sub_ca_certificates_oem = [
+                load_cert(CertPath.OEM_SUB_CA2_DER),
+                load_cert(CertPath.OEM_SUB_CA1_DER),
+            ]
+            root_ca_certificate_oem = CertPath.OEM_ROOT_DER
```
tropxy, line 686: `load_cert` already raises an exception if the certificates cant be fetched or is there an error

So, I would do the following:
```python
try:
    sub_ca_certificates_oem = [
        load_cert(CertPath.OEM_SUB_CA2_DER),
        load_cert(CertPath.OEM_SUB_CA1_DER),
    ]
    root_ca_certificate_oem = load_cert(CertPath.OEM_ROOT_DER)
except (FileNotFoundError, IOError):
    pass
```

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -577,61 +591,120 @@ async def process_message(
...
+        sub_ca_certificates_oem = None
+        root_ca_certificate_oem = None
+
+        try:
+            sub_ca_certificates_oem = [
+                load_cert(CertPath.OEM_SUB_CA2_DER),
+                load_cert(CertPath.OEM_SUB_CA1_DER),
+            ]
+            root_ca_certificate_oem = load_cert(CertPath.OEM_ROOT_DER)
+        except:  # noqa
```
tropxy, line 684: add the exceptions that it can raise so it wont complaint:
`except (FileNotFoundError, IOError)`

tropxy (changes requested):

tropxy (approved):

Issue #54, opened 2022-05-27
Handle rejection of authorization in 15118-2


danielgordon-switch-ev:
Correctly handle the case where the AuthorizationReq is rejected, in the PaymentDetails and Authorization states.  This should also include handling the case where the authorization is Ongoing (returns early to comply with ISO 15118 timeouts, but is continuing to be processed, and will be requested again).

Per @shalinnijel2 :

If Authorization is rejected we need to call:

            self.stop_state_machine(
                "Authorization was rejected",
                message,
                ResponseCode.FAILED,
            )

'message' above is the incoming AuthorizationReq message. stop_state_machine() would build the appropriate response message - in this case AuthorizationRes message with ResponseCode set to ResponseCode.FAILED and also set the next_state to Terminate.
One thing to keep in mind here is the failed response code in AuthorizationRes could be FAILED or as per V2G2-475, be FAILED_ChallengeInvalid - but I think that will be better handled under a separate ticket

Per @tropxy :

in case of Ongoing or Accepted we go to the Authorization state.
in case of rejected then we can answer with an Error code.

we also need to save the state of the authorization in the session here, I believe, because if the auth is successful in the PaymentDetailsReq, then we wont need to request it in the Authorization state and we can move forward. Unless, the ocpp client holds the authorization response and answers immediately if we do another request for the same eMAID.

if the answer "accepted" was not yet received, then we will keep requesting until the timeout (60s)

PR #83, opened 2022-07-15
feat: plug and charge authorization, basic happy path

danielgordon-switch-ev:
This implements the basic flow for Plug and Charge authorization.

Still to be done:
- handling the Ongoing case (managing around the short timeouts required by 15118)
- could probably use more testing of the technicalities of certificate handling
- accommodating better for other authorization cases (probably not for this PR)

Note that this prepares the certificate chain and hash data in the form in which OCPP will require them.

I've left some outstanding questions in TODOs for the time being as well -- let's either resolve or make issues for these.


iso15118/secc/states/iso15118_2_states.py
```diff
@@ -765,6 +768,49 @@ class PaymentDetails(StateSECC):
...
+        """Return a string of certificates in PEM form concatenated together."""
+        if certificate_chain is None:
+            return None
+
+        return "".join(
+            [
+                certificate_to_pem_string(certificate)
+                for certificate in self._all_certificates_from_chain(certificate_chain)
+            ]
+        )
```
HugoJP1, line 808: Think it would be good to add some unit tests for each of these methods to verify the behaviour.

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -791,24 +837,57 @@ async def process_message(
             # TODO Either an MO Root certificate or a V2G Root certificate
             #      could be used to verify, need to be flexible with regards
             #      to the PKI that is used.
-            verify_certs(leaf_cert, sub_ca_certs, CertPath.MO_ROOT_DER)
-
-            # TODO Check if EMAID has correct syntax
+            verify_certs(
+                leaf_cert, sub_ca_certs, self._mobility_operator_root_cert_path()
+            )
```
HugoJP1, line 842: Will add a ticket noting that the verification needs to be more dynamic than this.

iso15118/shared/security.py
```diff
@@ -1117,6 +1117,69 @@ def decrypt_priv_key(
...
+    public_key = certificate.public_key()
+    public_key_bytes = public_key.public_bytes(
+        encoding=Encoding.X962,
+        format=PublicFormat.UncompressedPoint,
+    )
+    # TODO: is this correct?
+    # gives a name like:
+    # 'DC=MO,C=DE,O=Keysight Technologies,CN=PKI-1_CRT_MO_SUB2_VALID'
+    # Do we need this whole thing or only, say, O?
+    distinguished_name = certificate.issuer.rfc4514_string().encode()
```
tropxy, line 1146: that is correct. The distinguisher name is the issuer:
https://www.ibm.com/docs/en/i/7.2?topic=concepts-distinguished-name
https://www.ibm.com/docs/en/ibm-mq/7.5?topic=certificates-distinguished-names

you can compare it when you extract the issuer, check here the command:
https://www.notion.so/switchev/Certificates-944733c8b468493a9e0eeace3b772d6a

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -791,24 +837,57 @@ async def process_message(
...
 
+            # TODO Check if EMAID has correct syntax -- is this accomplished by the
+            # constrained type in `datatypes.py`?
+            self.comm_session.emaid = payment_details_req.emaid
```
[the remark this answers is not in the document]
tropxy: yes, the syntax is checked by the type in dataypes

iso15118/shared/security.py
```diff
@@ -1117,6 +1117,69 @@ def decrypt_priv_key(
...
+
+    Args:
+        certificate: The certificate in binary (DER) form.
+
+    Returns:
+        The same certificate expressed as a PEM-format string.
+    """
+    certificate = load_der_x509_certificate(certificate)
+    pem_bytes = certificate.public_bytes(encoding=Encoding.PEM)
+    return pem_bytes.decode()
```
tropxy, line 1180: can we just add a note that the conversion is done because it is defined by the OCPP that it requires the chain to be pem encoded?

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -765,6 +768,49 @@ class PaymentDetails(StateSECC):
...
+        return "".join(
+            [
+                certificate_to_pem_string(certificate)
+                for certificate in self._all_certificates_from_chain(certificate_chain)
+            ]
+        )
+
+    def _mobility_operator_root_cert_path(self) -> str:
+        """Return the path to the MO root.  Included to be patched in tests."""
+        return CertPath.MO_ROOT_DER
```
tropxy, line 812: do we really need this?
cant we do the patching like I did here in pyslac?
https://github.com/SwitchEV/pyslac/blob/master/tests/test_slac_session.py#L95

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -791,24 +837,57 @@ async def process_message(
...
 
-            self.create_next_message(
-                Authorization,
-                payment_details_res,
-                Timeouts.V2G_SECC_SEQUENCE_TIMEOUT,
-                Namespace.ISO_V2_MSG_DEF,
+            authorization_result = (
+                await self.comm_session.evse_controller.is_authorized(
+                    id_token=payment_details_req.emaid,
+                    id_token_type=self.comm_session.selected_auth_option,
```
tropxy, line 859: I am not sure this is what we want...
selected_auth_option will get a value out of this:
```python
class AuthEnum(str, Enum):
    """
    The enum values for the authorisation options differ between DIN SPEC 70121,
    ISO 15118-2 and ISO 15118-20. This enumeration helps to unify different values.

    The default value for the enum members (EIM and PNC) are the ones from
    ISO 15118-20. They are used in the session variables and evcc/secc settings.

    For the ISO 15118-2 messages (see class AuthOptions), we add enum values
    EIM_V2 and PNC_V2, which provide the specific string value used in that
    standard.
    """

    EIM = "EIM"
    PNC = "PnC"
    EIM_V2 = "ExternalPayment"
    PNC_V2 = "Contract"
   ```
so basically Plug and Charge or External identification.

the id token type is ocpp specific and will be "eMAID" in the PnC case

I think the id token can be removed as an argument for the `is_authorized`. We can leave the job to derive what type it is to the ocpp client

File: iso15118/secc/states/iso15118_2_states.py [hunk 8]
```diff
@@ -791,24 +837,57 @@ async def process_message(
...
+                )
+
+                self.create_next_message(
+                    Authorization,
+                    payment_details_res,
+                    Timeouts.V2G_SECC_SEQUENCE_TIMEOUT,
+                    Namespace.ISO_V2_MSG_DEF,
+                )
+            else:
+                # TODO: handle ONGOING case, and REJECTED more fully
```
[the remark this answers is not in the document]
tropxy: isnt this part quite straight forward? in case of Ongoing or Accepted we go to the Authorization state.
in case of rejected than we can answer with an Error code.

we also need to save the state of the authorization in the session here, I believe, because if the auth is successful in the PaymentDetailsReq, then we wont need to request it in the Authorization state and we can move forward. Unless, the ocpp client holds the authorization response and answer immediately if we do another request for the same eMAID.

if the answer "accepted" was not yet received, then we will keep requesting until the timeout (60s)

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -930,9 +1009,17 @@ async def process_message(
...
-        ):
+        # TODO: how should we obtain/hold the token for EIM?  Other cases?
+        id_token = (
+            self.comm_session.emaid
+            if self.comm_session.selected_auth_option == AuthEnum.PNC_V2
+            else None
+        )
+        authorization_result = await self.comm_session.evse_controller.is_authorized(
+            id_token=id_token,
+            id_token_type=self.comm_session.selected_auth_option,
```
tropxy, line 1020: just again a note about the auth option may not be what we want

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -765,6 +768,49 @@ class PaymentDetails(StateSECC):
...
+            certificate_chain.certificate
+        ] + certificate_chain.sub_certificates.certificates
+
+    def _get_contract_certificate_hash_data(
+        self,
+        certificate_chain: Optional[CertificateChain],
+    ) -> Optional[List[Dict[str, str]]]:
+        """Return a list of hash data for a contract certificate chain."""
+        if certificate_chain is not None:
+            return None
```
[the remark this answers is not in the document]
tropxy: correct

iso15118/shared/security.py
```diff
@@ -1117,6 +1117,69 @@ def decrypt_priv_key(
...
+    issuer_name_hasher = Hash(certificate.signature_hash_algorithm)
+    issuer_name_hasher.update(distinguished_name)
+
+    return {
+        "hash_algorithm": hash_algorithm_for_ocpp,
+        "issuer_name_hash": issuer_name_hasher.finalize(),
+        "issuer_key_hash": public_key_hasher.finalize(),
+        "serial_number": serial_number,
+        # TODO: Populate with a real-world verification server
+        "responder_url": "https://www.example.com/",
```
tropxy, line 1165: I guess it will look something like this: "http://ocsp-qa.hubject.com:8080" (this is the QA env for the OCSP validation in Hubject.

tropxy: I did some research and this is what we need to do to extract the responder_url

```python
from cryptography.x509.oid import ExtensionOID, AuthorityInformationAccessOID

def get_ocsp_url(cert):
    auth_inf_access = cert.extensions.get_extension_for_oid(ExtensionOID.AUTHORITY_INFORMATION_ACCESS).value
    ocsps = [access_descriptor for access_descriptor in auth_inf_access if access_descriptor.access_method == AuthorityInformationAccessOID.OCSP]
    if not ocsps:
        raise Exception(f'no ocsp server entry in Authority Information Access extension field')
    return ocsps[0].access_location.value
```
If the Authority Information Access is not found, the method raises a `ExtensionNotFound` error

References: 
https://stackoverflow.com/questions/64436317/how-to-check-ocsp-client-certificate-revocation-using-python-requests-library

https://cryptography.io/en/latest/x509/reference/?highlight=AuthorityInformationAccessOID.CA_ISSUERS#cryptography.x509.Extensions.get_extension_for_oid

https://cryptography.io/en/latest/x509/reference/?highlight=AuthorityInformationAccessOID.OCSP#cryptography.x509.AccessDescription

https://cryptography.io/en/latest/x509/reference/?highlight=AuthorityInformationAccessOID.OCSP#cryptography.x509.oid.AuthorityInformationAccessOID

tropxy: Also, to get the hashed info for the OCSP request we can do this:
```python
from cryptography.x509.ocsp import OCSPRequestBuilder

def get_oscp_request(cert: bytes, issuer_cert: bytes):
    """
    cert: The certificate in DER format
    issuer_cert: The certificate used to sign the `cert` certificate, also called the issuer certificate
    """
    cert = load_der_x509_certificate(cert)
    issuer_cert = load_der_x509_certificate(issuer_cert)
    builder = OCSPRequestBuilder().add_certificate(cert, issuer_cert, cert.signature_hash_algorithm)
    return builder.build()

ocsp_request = get_oscp_request(certificate, issuer_cert)
```
The data can then be accessed through the returned object, for example
`ocsp_request.issuer_key_hash`.

Not sure if this is better as we need to load the issuer certificate, but makes it easier then to extract the needed data.

Reference:
https://cryptography.io/en/latest/x509/ocsp/#cryptography.x509.ocsp.OCSPRequest

tropxy (changes requested):
added some suggestions and food for thought

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: @HugoJP1 Have you done this, or shall I pick it back up?

iso15118/shared/security.py
danielgordon-switch-ev: Certainly -- I've mentioned it elsewhere as well but it can only be helpful to include it here.

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: Quite possibly -- I added this because it was the quickest way I saw to get the tests passing.

However, might it be better to leave this in, since it doesn't complicate things now too much, and will be useful in case we have certificates of multiple mobility operators locally (in which case we'd need to access the correct one, rather than just loading from a hardcoded path)?

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: I'm not sure I follow -- without this how would is_authorized derive the type?  For example a 14-character eMAID could also be an ISO 14443 (7 bytes in hexadecimal form) token.

danielgordon-switch-ev:
For some reason I can't respond directly to some of your comments.

The OCSP URL stuff looks good to me -- thanks for digging this up.

On handling authorization, I'd prefer to do this as a separate follow-up PR -- may do some minor refactoring relating to is_authorized as well.

Agreed that we may not want to use the OCSPRequest object if it requires us to have the root certificate.

iso15118/shared/security.py
tropxy: sounds good, I am all in favour of comprehensive comments ;)

iso15118/secc/states/iso15118_2_states.py
tropxy: hmm, I understand your point of view, but with patching we can achieve the same thing...you can decide what to return in different tests. I just dont feel that is right to leave methods on the way of "real code" specially given that there is a solution to avoid it and keep functional code and tests in separate.

iso15118/secc/states/iso15118_2_states.py
tropxy: well, is_authorized here in the iso15118 world will either be nothing or the eMAID and possibly can also be the MAC address, but this use case is still  bit undefined and I think the authentication using the MAC address wouldnt be triggered by the iso15118 either.

So, in general in the OCPP if the authorization is coming from 15118, then we know it is an eMAID.
But ok, let's not complicate and try to derivate types. So, here what we need to do is in case of PnC we need to add the right type that the `iso15118_authorization`method would accept

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -791,24 +837,57 @@ async def process_message(
...
-            self.create_next_message(
-                Authorization,
-                payment_details_res,
-                Timeouts.V2G_SECC_SEQUENCE_TIMEOUT,
-                Namespace.ISO_V2_MSG_DEF,
+            authorization_result = (
+                await self.comm_session.evse_controller.is_authorized(
+                    id_token=payment_details_req.emaid,
+                    id_token_type=self.comm_session.selected_auth_option,
+                    certificate_chain=pem_certificate_chain,
```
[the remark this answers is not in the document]
danielgordon-switch-ev: Very interesting...`certificate` is not required in the OCPP AuthorizeRequest in any case, since it would only be used for contract certificates, and the OCSPRequestDataType only requires the hashes and the serial number.

Do we have a copy of this document, or is it publicly available?  Could you point me to it if so?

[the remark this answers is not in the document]
danielgordon-switch-ev: After giving this more thought, is it possible that providers other than Hubject might want the certificate chain for reasons of their own, even if Hubject doesn't?  It might be simpler to leave it in for now until we know more.

iso15118/secc/states/iso15118_2_states.py
tropxy: I understand your point that the 15118 world shall not clash with our MQTT private world, but if we name the attribute id_token_type then we shall be consistent...maybe we can create an Enum in iso15118 code and for now just add the type "eMAID" in it

iso15118/shared/security.py
danielgordon-switch-ev: Trying this with some of the sample certificates, they don't seem to have an Authority Information Access extension.  For now I'm just falling back to the sample value we currently have, but that probably just pushes the error elsewhere.  So maybe it's better to revise this to fail on these exceptions, since it won't be possible to verify the certificate.

iso15118/shared/security.py
```diff
@@ -1112,6 +1115,128 @@ def decrypt_priv_key(
...
+    public_key_hasher.update(public_key_bytes)
+    issuer_name_hasher = Hash(certificate.signature_hash_algorithm)
+    issuer_name_hasher.update(distinguished_name_bytes)
+
+    try:
+        responder_url = get_ocsp_url_for_certificate(certificate)
+    except (ExtensionNotFound, OCSPServerNotFoundError):
+        # TODO: This may just result in failure down the road.
+        # Should we let this fail on these exceptions, or is there
+        # another way to try to get a responder_url?
```
danielgordon-switch-ev, line 1177: @tropxy What do you think -- is it an unrecoverable failure if the certificate does not provide a URL for an OCSP server, or is there something else we could try in this case?

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -857,24 +904,59 @@ async def process_message(
             # TODO Either an MO Root certificate or a V2G Root certificate
             #      could be used to verify, need to be flexible with regards
             #      to the PKI that is used.
-            verify_certs(leaf_cert, sub_ca_certs, load_cert(CertPath.MO_ROOT_DER))
+            with open(self._mobility_operator_root_cert_path(), "rb") as root_cert_file:
+                root_cert = root_cert_file.read()
```
danielgordon-switch-ev, line 908: Will change this to also use load_cert.

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: Done now.

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: There's more to this than patching -- at some point we may not have only one mobility root certificate, right?  Different mobility operators may ultimately derive from different roots, and we might need to have them both on the same charging station (or is this wrong?)

I probably wouldn't have added this just for that purpose (yet), but now that it's in it seems to me fine to leave it there.

If you insist, I can move to patching for tests and get rid of this.

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: As mentioned, have resolved this.

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -857,24 +904,57 @@ async def process_message(
             # TODO Either an MO Root certificate or a V2G Root certificate
             #      could be used to verify, need to be flexible with regards
             #      to the PKI that is used.
-            verify_certs(leaf_cert, sub_ca_certs, load_cert(CertPath.MO_ROOT_DER))
-
-            # TODO Check if EMAID has correct syntax
+            root_cert_path = self._mobility_operator_root_cert_path()
+            verify_certs(leaf_cert, sub_ca_certs, load_cert(root_cert_path))
```
tropxy, line 908: we dont have a guarantee that the MO root will be available in the station. So, we may actually call a evse_controller to get a True/False result for the question "verify contract certificate chain?"

iso15118/shared/security.py
```diff
@@ -1112,6 +1115,128 @@ def decrypt_priv_key(
...
+    distinguished_name_bytes = certificate.issuer.public_bytes()
+    serial_number = certificate.serial_number
+
+    # Convert to the naming used in OCPP.
+    hash_algorithm_for_ocpp = certificate.signature_hash_algorithm.name.upper()
+    if hash_algorithm_for_ocpp not in {"SHA256", "SHA384", "SHA512"}:
+        raise CertAttributeError("Unknown hash algorithm")
+
+    public_key_hasher = Hash(certificate.signature_hash_algorithm)
+    public_key_hasher.update(public_key_bytes)
```
tropxy, line 1168: the public key we should extract is the public key bytes from the Issuer certificate and not from the Contract Certificate

iso15118/shared/security.py
```diff
@@ -1112,6 +1115,128 @@ def decrypt_priv_key(
...
+        certificate: The certificate in binary (DER) form.
+
+    Returns:
+        The same certificate expressed as a PEM-format string.
+    """
+    certificate = load_der_x509_certificate(certificate)
+    # This conversion is done because OCPP requires that the certificate chain be
+    # PEM-encoded.
+    pem_bytes = certificate.public_bytes(encoding=Encoding.PEM)
+    return pem_bytes.decode()
```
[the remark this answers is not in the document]
danielgordon-switch-ev: Yes, let's use that.

iso15118/secc/states/iso15118_2_states.py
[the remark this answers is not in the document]
danielgordon-switch-ev: Attached this to the existing https://github.com/SwitchEV/iso15118/issues/54

iso15118/shared/security.py
danielgordon-switch-ev: Did this.  Additionally, may still move to the OCSPRequestBuilder here.

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: https://github.com/SwitchEV/iso15118/issues/94

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -864,24 +871,66 @@ async def process_message(
...
+
+            if authorization_result == AuthorizationStatus.ACCEPTED:
+
+                payment_details_res = PaymentDetailsRes(
+                    response_code=ResponseCode.OK,
+                    gen_challenge=get_random_bytes(16),
+                    evse_timestamp=time.time(),
+                )
+
+                # TODO: Should the next message really be Authorization?  If so why?
```
tropxy, line 912: If the Auth status is Accepted or Ongoing, yes, the next step shall be the Authorization class, as that is the message the EV will send after reception of the PaymentDetailsRes

iso15118/shared/security.py
danielgordon-switch-ev: https://github.com/SwitchEV/iso15118/issues/96

iso15118/secc/states/iso15118_2_states.py
danielgordon-switch-ev: I think this falls under #94 .

iso15118/shared/security.py
```diff
@@ -1125,6 +1130,218 @@ def decrypt_priv_key(
...
+    if hash_algorithm_for_ocpp not in {"SHA256", "SHA384", "SHA512"}:
+        raise CertAttributeError("Unknown hash algorithm")
+
+    try:
+        responder_url = get_ocsp_url_for_certificate(certificate)
+    except (ExtensionNotFound, OCSPServerNotFoundError):
+        # TODO GitHub#96: This may just result in failure down the road.
+        # Should we let this fail on these exceptions, or is there
+        # another way to try to get a responder_url?
+        responder_url = "https://www.example.com/"
```
tropxy, line 1174: if we cant get the OCSP url, then the OCSP call cant happen, so we may just drop everything and end as a failure, ie authorization has failed

tropxy (changes requested):

iso15118/shared/security.py
danielgordon-switch-ev: This might be the best way we have to resolve #96 then.

tropxy (approved):
LGTM

PR #85, opened 2022-07-28
fix: EXI grammar violation for failed response.

shalinnijel2:
Failed response generated for CertificateInstallationRes was incorrect.
Short summary of changes:
- Certificate field in the generated CertificateInstallationRes was violating EXI grammar. The field can't be empty - added a single char string to the field.
- Id field in EMAID is not optional - fixed.


iso15118/secc/failed_responses.py
```diff
@@ -599,16 +599,16 @@ def init_failed_responses_iso_v20() -> dict:
...
                 signed_installation_data=SignedInstallationData(
                     contract_cert_chain=CertificateChainV20(
-                        certificate=bytes(0),
-                        sub_certificates=SubCertificates(certificates=[bytes(0)]),
+                        certificate=b"_",
+                        sub_certificates=SubCertificates(certificates=[b"_"]),
                     ),
                     ecdh_curve=ECDHCurve.x448,
-                    dh_public_key=bytes(0),
+                    dh_public_key=b"_",
```
tropxy, line 609: Let's do like Hubject and do this:
```python
In [11]: bytes(1)
Out[11]: b'\x00'
```

tropxy (approved):

PR #87, opened 2022-07-29
fix: wrong message parameters will return FAILED_WRONG_CHARGE_PARAMETER

ikaratass:
Fixed [AB#2522](https://dev.azure.com/switch-ev/09ab882d-ca4b-4b5b-b5b8-80603c12e5b2/_workitems/edit/2522)

For the ChargeParameterDiscoveryReq, it is supposed to be return FAILED_WRONG_CHARGE_PARAMETER when a wrong parameter received.

### DIFF
--- a/iso15118/shared/exi_codec.py
+++ b/iso15118/shared/exi_codec.py
@@ -6,7 +6,11 @@ from typing import Union
 
 from pydantic import ValidationError
 
-from iso15118.shared.exceptions import EXIDecodingError, EXIEncodingError
+from iso15118.shared.exceptions import (
+    EXIDecodingError,
+    EXIEncodingError,
+    V2GMessageValidationError,
+)
 from iso15118.shared.exificient_exi_codec import ExificientEXICodec
 from iso15118.shared.iexi_codec import IEXICodec
 from iso15118.shared.messages import BaseModel
@@ -14,8 +18,11 @@ from iso15118.shared.messages.app_protocol import (
     SupportedAppProtocolReq,
     SupportedAppProtocolRes,
 )
+from iso15118.shared.messages.din_spec.body import get_msg_type as get_msg_type_dinspec
 from iso15118.shared.messages.din_spec.msgdef import V2GMessage as V2GMessageDINSPEC
 from iso15118.shared.messages.enums import Namespace
+from iso15118.shared.messages.iso15118_2.body import get_msg_type
+from iso15118.shared.messages.iso15118_2.datatypes import ResponseCode
 from iso15118.shared.messages.iso15118_2.msgdef import V2GMessage as V2GMessageV2
 from iso15118.shared.messages.iso15118_20.ac import (
     ACChargeLoopReq,
@@ -379,10 +386,28 @@ class EXI:
 
             raise EXIDecodingError("Can't identify protocol to use for decoding")
         except ValidationError as exc:
-            raise EXIDecodingError(
-                f"Error parsing the decoded EXI into a Pydantic class: {exc}. "
-                f"\n\nDecoded dict: {decoded_dict}"
+            if namespace == Namespace.ISO_V2_MSG_DEF:
+                msg_name = next(iter(decoded_dict["V2G_Message"]["Body"]))
+                msg_type = get_msg_type(msg_name)
+            elif namespace == Namespace.DIN_MSG_DEF:
+                msg_name = next(iter(decoded_dict["V2G_Message"]["Body"]))
+                msg_type = get_msg_type_dinspec(msg_name)
+            elif namespace.startswith(Namespace.ISO_V20_BASE):
+                msg_type = msg_class
+            elif namespace == Namespace.SAP:
+                if "supportedAppProtocolReq" in decoded_dict:
+                    msg_type = SupportedAppProtocolReq
+                elif "supportedAppProtocolRes" in decoded_dict:
+                    msg_type = SupportedAppProtocolRes
+
+            raise V2GMessageValidationError(
+                f"Validation error: {exc}. \n\nDecoded dict: " f"{decoded_dict}",
+                ResponseCode.FAILED,
+                msg_type,
             ) from exc
+
+        except V2GMessageValidationError as exc:
+            raise exc
         except EXIDecodingError as exc:
             raise EXIDecodingError(
                 f"EXI decoding error: {exc}. \n\nDecoded dict: " f"{decoded_dict}"
--- a/iso15118/shared/messages/din_spec/body.py
+++ b/iso15118/shared/messages/din_spec/body.py
@@ -4,6 +4,7 @@ from typing import Optional, Tuple, Type
 
 from pydantic import Field, root_validator, validator
 
+from iso15118.shared.exceptions import V2GMessageValidationError
 from iso15118.shared.messages import BaseModel
 from iso15118.shared.messages.datatypes import (
     DCEVSEChargeParameter,
@@ -234,11 +235,6 @@ class ChargeParameterDiscoveryReq(BodyBase):
 
         Pydantic validators are "class methods",
         see https://pydantic-docs.helpmanual.io/usage/validators/
-        TODO We need to actually send FAILED_WrongChargeParameter or
-             FAILED_WrongEnergyTransferMode if the wrong parameter set is
-             provided, one or multiple parameters can not be interpreted
-             (see [V2G2-477]). Need to check how to not just bury
-             that information in a pydantic validation error.
         """
         # pylint: disable=no-self-argument
         # pylint: disable=no-self-use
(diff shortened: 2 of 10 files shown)


ikaratass:
https://github.com/SwitchEV/josev_pro/issues/90

tropxy:
there is no description in the PR of what it is supposed to fix or a reference to a sprint ticket.

"wrong message parameters will return FAILED_WRONG_CHARGE_PARAMETER" doesnt tell me what is the problem.
can you elaborate on that?

iso15118/shared/exi_codec.py
```diff
@@ -378,11 +380,18 @@ def from_exi(
                 return msg_class.parse_obj(msg_dict)
 
             raise EXIDecodingError("Can't identify protocol to use for decoding")
-        except ValidationError as exc:
-            raise EXIDecodingError(
-                f"Error parsing the decoded EXI into a Pydantic class: {exc}. "
-                f"\n\nDecoded dict: {decoded_dict}"
-            ) from exc
+        except V2GMessageValidationError as exc:
```
tropxy, line 383: I dont think you can just substitute the Pydantic error `ValidationError` for the `V2GMessageValidationError` as there are errors that Pydantic will throw as `ValidationError` because we tried to parse an integer into a field that was supposed to be a str for example...

iso15118/shared/messages/din_spec/body.py
```diff
@@ -249,15 +250,19 @@ def validate_requested_energy_mode(cls, values):
...
             )
         if ("AC_" in requested_energy_mode and dc_params) or (
             "DC_" in requested_energy_mode and ac_params
         ):
-            raise ValueError(
+            raise V2GMessageValidationError(
                 "Wrong charge parameters for requested energy "
-                f"transfer mode {requested_energy_mode}"
+                f"transfer mode {requested_energy_mode}",
+                ResponseCode.FAILED_WRONG_ENERGY_TRANSFER_MODE,
```
tropxy, line 264: here should be `FAILED_WRONG_CHARGE_PARAMETER`

iso15118/shared/messages/iso15118_2/body.py
```diff
@@ -268,9 +268,11 @@ def requested_energy_mode_must_match_charge_parameter(cls, values):
         if ("AC_" in requested_energy_mode and dc_params) or (
             "DC_" in requested_energy_mode and ac_params
         ):
-            raise ValueError(
+            raise V2GMessageValidationError(
                 "Wrong charge parameters for requested energy "
-                f"transfer mode {requested_energy_mode}"
+                f"transfer mode {requested_energy_mode}",
+                ResponseCode.FAILED_WRONG_CHARGE_PARAMETER,
+                None,
```
tropxy, line 275: to make it complete I guess we also need to check that the `requested_energy_mode´ is part of the `EnergyTransferModeEnum` values, otherwise we shall raise a `FAILED_WRONG_ENERGY_TRANSFER_MODE`

tropxy (changes requested):

iso15118/shared/exi_codec.py
ikaratass: Done

iso15118/shared/messages/din_spec/body.py
ikaratass: Done

iso15118/shared/messages/iso15118_2/body.py
ikaratass: done

iso15118/shared/exi_codec.py
```diff
@@ -383,6 +387,18 @@ def from_exi(
...
+                exc.message = V2GMessageV2.construct_skip_validation(
+                    **decoded_dict["V2G_Message"]
+                )
+            elif namespace == Namespace.DIN_MSG_DEF:
+                exc.message = V2GMessageDINSPEC.construct_skip_validation(
+                    **decoded_dict["V2G_Message"]
+                )
+            elif namespace.startswith(Namespace.ISO_V20_BASE):
+                exc.message = msg_class.construct(**msg_dict)
+            raise exc
```
tropxy, line 401: this I guess is solving part of the problem but not all of it.
The Pydantic `parse_obj` method will try to construct a python class based on the dictionary message we pass to it...

Now, if that parsing fails, a `ValidationError` is raised (https://pydantic-docs.helpmanual.io/usage/validators/) and currently it seems the way we deal with it is this:
```python
raise EXIDecodingError(
                f"Error parsing the decoded EXI into a Pydantic class: {exc}. "
                f"\n\nDecoded dict: {decoded_dict}"
            ) from exc
```

so maybe here we need to raise a `V2GMessageValidationError` instead?

tropxy (changes requested):

iso15118/shared/messages/iso15118_2/body.py
```diff
@@ -268,10 +268,20 @@ def requested_energy_mode_must_match_charge_parameter(cls, values):
...
             )
+        if requested_energy_mode not in EnergyTransferModeEnum.__members__.values():
+            raise V2GMessageValidationError(
+                "Wrong value for requested energy "
+                f"transfer mode {requested_energy_mode}",
+                ResponseCode.FAILED_WRONG_ENERGY_TRANSFER_MODE,
```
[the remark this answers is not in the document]
ikaratass: Done

iso15118/shared/messages/iso15118_2/body.py
```diff
@@ -268,10 +268,20 @@ def requested_energy_mode_must_match_charge_parameter(cls, values):
         if ("AC_" in requested_energy_mode and dc_params) or (
             "DC_" in requested_energy_mode and ac_params
         ):
-            raise ValueError(
+            raise V2GMessageValidationError(
                 "Wrong charge parameters for requested energy "
-                f"transfer mode {requested_energy_mode}"
+                f"transfer mode {requested_energy_mode}",
+                ResponseCode.FAILED_WRONG_CHARGE_PARAMETER,
+                None,
```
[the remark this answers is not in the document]
ikaratass: Done

iso15118/shared/exi_codec.py
ikaratass: Done.

tropxy (approved):

iso15118/shared/messages/iso15118_2/body.py
```diff
@@ -705,7 +712,7 @@ def get_message_name(self) -> str:
         """Returns the name of the one V2GMessage that is set for Body."""
         for k in self.__dict__.keys():
             if getattr(self, k):
-                return str(getattr(self, k))
+                return str(k)
```
tropxy, line 715: have you tested this? because this change will break the intention of this code...it will no longer print the class name but the key of the dictionary.
The dictionary of this class is like:
```python
{'current_demand_req': None, 'current_demand_res': CurrentDemandRes(....), 'metering_receipt_req*: None}`
````
And if you return `str(k)` will return 
`current_demand_res`

**Review by tropxy — CHANGES_REQUESTED** [review]

File: iso15118/shared/messages/iso15118_2/body.py (hunk 7 above)
Comment (ikaratass) [reply iso15118/shared/messages/iso15118_2/body.py:715]: Yes, it is tested and actually, this is what we want, you also can understand from the comment above and method name. Additionally, the return value will be used to find the message type in `get_msg_type` method which has the message names table.

File: iso15118/shared/messages/din_spec/body.py [hunk 8]
```diff
@@ -234,7 +237,7 @@ def validate_requested_energy_mode(cls, values):
 
         Pydantic validators are "class methods",
         see https://pydantic-docs.helpmanual.io/usage/validators/
-        TODO We need to actually send FAILED_WrongChargeParameter or
+        We need to actually send FAILED_WrongChargeParameter or
```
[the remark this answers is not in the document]
ikaratass: Done

iso15118/shared/messages/iso15118_2/body.py
tropxy: get_msg_type is defined as:
```python
def get_msg_type(msg_name: str) -> Optional[Type[BodyBase]]:
    """
    Returns the message type corresponding to the message name provided, or
    None if not match is found.

    Args:
        msg_name: The name of the message (e.g. SessionSetupReq)

    Returns: The message type corresponding to the given message name
    """
    msg_dict = {
        "SessionSetupReq": SessionSetupReq,
        "SessionSetupRes": SessionSetupRes,
        "ServiceDiscoveryReq": ServiceDiscoveryReq,
        "ServiceDiscoveryRes": ServiceDiscoveryRes,
        "ServiceDetailReq": ServiceDetailReq,
        "ServiceDetailRes": ServiceDetailRes,
        "PaymentServiceSelectionReq": PaymentServiceSelectionReq,
        "PaymentServiceSelectionRes": PaymentServiceSelectionRes,
        "CertificateInstallationReq": CertificateInstallationReq,
        "CertificateInstallationRes": CertificateInstallationRes,
        "PaymentDetailsReq": PaymentDetailsReq,
        "PaymentDetailsRes": PaymentDetailsRes,
        "AuthorizationReq": AuthorizationReq,
        "AuthorizationRes": AuthorizationRes,
        "CableCheckReq": CableCheckReq,
        "CableCheckRes": CableCheckRes,
        "PreChargeReq": PreChargeReq,
        "PreChargeRes": PreChargeRes,
        "ChargeParameterDiscoveryReq": ChargeParameterDiscoveryReq,
        "ChargeParameterDiscoveryRes": ChargeParameterDiscoveryRes,
        "PowerDeliveryReq": PowerDeliveryReq,
        "PowerDeliveryRes": PowerDeliveryRes,
        "ChargingStatusReq": ChargingStatusReq,
        "ChargingStatusRes": ChargingStatusRes,
        "CurrentDemandReq": CurrentDemandReq,
        "CurrentDemandRes": CurrentDemandRes,
        "MeteringReceiptReq": MeteringReceiptReq,
        "MeteringReceiptRes": MeteringReceiptRes,
        "WeldingDetectionReq": WeldingDetectionReq,
        "WeldingDetectionRes": WeldingDetectionRes,
        "SessionStopReq": SessionStopReq,
        "SessionStopRes": SessionStopRes,
    }
    return msg_dict.get(msg_name, None)
```

so, the key for checking it must be something like "SessionStopReq", but according with the mod you did, the key will be `session_stop_req` instead. Or am I missing something?

tropxy: any update on this @ikaratass

ikaratass: You were right about this issue, the construct method of pydantic was creating different objects rather than expected, so i changed my implementation completely removing object creation without validation. we don't do that anymore. So this part of the code is reverted.

tropxy:
just rebase the branch and then I think is good to go

tropxy (approved):

PR #88, opened 2022-08-02
Feat: support multiple passwords for private keys (AB#2546)

shalinnijel2:
Summary of the changes:
- create_certs.sh now writes the password used during certs generation into 5 separate files - seccLeafPassword.txt, oemLeafPassword.txt, contractLeafPassword.text, cpsLeafPassword.txt - this will help configure different passwords for different private keys. Very recently a private key received from a third party had no password set while the seccLeaf.key had password set. So a single password for all private keys isn't sufficient anymore.
- The respective password files are explicitly passed in while attempting to load a private key
- CERTS_GENERAL_PRIVATE_KEY_PASS_PATH env variable has been removed. Readme has been updated to reflect this.
- PKI_PATH is read as a "path" type in settings - helps fix in inconvenience where user has to remember to put a "/" at the end of the path otherwise. Source security.py has been updated to reflect this.


iso15118/shared/security.py
```diff
@@ -1162,25 +1175,54 @@ class KeyPath(str, Enum):
...
+    OEM_LEAF_PEM = os.path.join(PKI_PATH, "iso15118_2/private_keys/oemLeaf.key")
+    OEM_SUB_CA2_PEM = os.path.join(PKI_PATH, "iso15118_2/private_keys/oemSubCA2.key")
+    OEM_SUB_CA1_PEM = os.path.join(PKI_PATH, "iso15118_2/private_keys/oemSubCA1.key")
+    OEM_ROOT_PEM = os.path.join(PKI_PATH, "iso15118_2/private_keys/oemRootCA.key")
+
+
+class KeyPasswordPath(str, Enum):
+    """
+    Provides the path to private key passwords used for Plug & Charge. The encoding
+    format is indicated by the latter part of the enum name (_DER or _PEM)
```
tropxy, line 1207: the docstring is incorrect. the DER, PEM stuff is not part of this ;)

tropxy (changes requested):

iso15118/shared/security.py
shalinnijel2: Oops...missed while copy pasting and editing. Have changed now.

tropxy (approved):

PR #92, opened 2022-08-04
fix: missing parameters are added for ChargeParameterDiscoveryRes and PowerDeliveryRes

ikaratass:
[AB#2563](https://dev.azure.com/switch-ev/09ab882d-ca4b-4b5b-b5b8-80603c12e5b2/_workitems/edit/2563)

ac_evse_charge_parameter is added for ChargeParameterDiscoveryRes

ac_evse_status is added for PowerDeliveryRes


iso15118/secc/failed_responses.py
```diff
@@ -475,14 +481,31 @@ def init_failed_responses_iso_v2() -> dict:
...
+                ac_evse_status=ACEVSEStatus(
+                    notification_max_delay=0,
+                    evse_notification=EVSENotification.NONE,
+                    rcd=False,
+                ),
+                evse_nominal_voltage=PVEVSENominalVoltage(
+                    multiplier=0, value=400, unit=UnitSymbol.VOLTAGE
+                ),
+                evse_max_current=PVEVSEMaxCurrent(
+                    multiplier=0, value=32, unit=UnitSymbol.AMPERE
```
shalinnijel2, line 497: Do you want to set the value field to 0.
Just to avoid the case where the recipient ignores the response_code and still decide to use the response?

shalinnijel2 (changes requested):

iso15118/secc/failed_responses.py
ikaratass: Done

shalinnijel2 (approved):

PR #102, opened 2022-08-10
get_contactor_state has been added

ikaratass:
The contactor state will be checked before the close/open command in order to avoid sending an unnecessary message if it is already closed/opened.


ikaratass:
it could be done on josev pro without modification of iso15118.

PR #103, opened 2022-08-11
Feature/add secc configurable parameters

aadritG:
Feature added to allow usage of configuration files (.yaml) passed as argument for setting the SECC configuration parameters for the charging session. This allows the tester to keep separate SECC configurations for distinct testing scenarios, avoiding the need to change the source code each time.

example:
$ make  run-secc  config=conf_default.yaml

If no argument is passed. default configurations shall be selected from conf_default.yaml

The configuration file is read in iso15118/secc/main.py and the returned dictionary is passed to SimEVSEController.create() and config.load_envs(). During the charging session, If the parameters in the config file are existent and valid, it is loaded, else the default fallback value is loaded.

Note: 
Might need to do the following to install pyyaml,
$poetry  add  pyyaml

Additionally, the protocol list in secc_settings has been sorted.

- Aadrit Ghimire, Lucid Motors


tropxy:
hi, thanks for your contribution, could you rebase your branch?

aadritG:
Rebased

iso15118/secc/controller/simulator.py
```diff
@@ -908,3 +953,15 @@ async def get_15118_ev_certificate(
         ).decode("utf-8")
 
         return base64_encode_cert_install_res
+
+    def get_cfg_parameter(self, string, default):
+        try:
+                if self.config_param[string] is None:                      
+                        raise Exception("revert to default")
```
tropxy, line 960: there is no point on raising a general exception here. log as info and return the default

iso15118/secc/controller/simulator.py
```diff
@@ -908,3 +953,15 @@ async def get_15118_ev_certificate(
...
+                logger.error(f"No valid configuration for {string}, revert to default")
```
tropxy, line 961: the message should be in the same indentation level as the raise above, nevertheless as I said I think this should be a INFO level and not an ERROR one

iso15118/secc/main.py
```diff
@@ -1,21 +1,39 @@
...
-    sim_evse_controller = await SimEVSEController.create()
+
+    global args
+
+    default_config_file = "conf_default.yaml"
+
+    try:
+        with open(args[1]) as f:
+            config_param = yaml.load(f, Loader=yaml.FullLoader)
+        print("Loading configuration parameters from " + args[1])
```
tropxy, line 27: you can remove this line

iso15118/secc/main.py
```diff
@@ -1,21 +1,39 @@
...
+        logger.info(f"Loaded configuration parameters from: {args[1]}")
+    except:
```
tropxy, line 29: what is the specific exception that is raised? I would assume "FileNotFoundError"?

tropxy (changes requested):

iso15118/secc/controller/simulator.py
aadritG: Updated! Thanks for the suggestion

iso15118/secc/controller/simulator.py
aadritG: Updated as specified above

iso15118/secc/main.py
aadritG: Yes. Additionally, if no configfile is passed as argument, line 25 will result in an IndexError

> with open(args[1]) as f:

PR #104, opened 2022-08-12
Added evse-id to is_authorized() request for EIM auth

shalinnijel2:
Passes through the evse id in is_authorized(). EVSE ID could be beneficial to is_authorized() implementation when EIM is used.


shalinnijel2:
Closing this PR. EVSE ID could be fetched easily from the interface implementation with await self.get_evse_id() anyway.

PR #107, opened 2022-08-17
Feat/complete pnc auth

### LINKED ISSUE #54: Handle rejection of authorization in 15118-2

tropxy:
Fix #54

### DIFF
--- a/iso15118/secc/controller/simulator.py
+++ b/iso15118/secc/controller/simulator.py
@@ -49,8 +49,8 @@ from iso15118.shared.messages.din_spec.datatypes import (
     SAScheduleTupleEntry as SAScheduleTupleEntryDINSPEC,
 )
 from iso15118.shared.messages.enums import (
-    AuthorizationTokenType,
     AuthorizationStatus,
+    AuthorizationTokenType,
     Contactor,
     EnergyTransferModeEnum,
     IsolationLevel,
--- a/iso15118/secc/controller/simulator.py
+++ b/iso15118/secc/controller/simulator.py
@@ -49,7 +49,7 @@ from iso15118.shared.messages.din_spec.datatypes import (
     SAScheduleTupleEntry as SAScheduleTupleEntryDINSPEC,
 )
 from iso15118.shared.messages.enums import (
-    AuthEnum,
+    AuthorizationTokenType,
     AuthorizationStatus,
     Contactor,
     EnergyTransferModeEnum,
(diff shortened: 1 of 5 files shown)


iso15118/secc/controller/simulator.py
```diff
@@ -423,12 +423,12 @@ async def get_energy_service_list(self) -> ServiceList:
...
         self,
         id_token: Optional[str] = None,
-        id_token_type: Optional[AuthEnum] = None,
+        id_token_type: Optional[AuthorizationTokenType] = None,
         certificate_chain: Optional[bytes] = None,
         hash_data: Optional[List[Dict[str, str]]] = None,
     ) -> AuthorizationStatus:
         """Overrides EVSEControllerInterface.is_authorized()."""
-        return AuthorizationStatus.ACCEPTED
+        return AuthorizationStatus.ONGOING
```
shalinnijel2, line 431: Should this be set back to 
```
return AuthorizationStatus.ACCEPTED
```

shalinnijel2 (changes requested):
One minor change...rest lgtm...

iso15118/secc/controller/simulator.py
tropxy: yes, thanks ;)

tropxy: done

shalinnijel2 (approved):

PR #109, opened 2022-08-18
fix: cleanup template dockerfile

mdwcrft:
Small clean up of the template dockerfile to remove some unneeded steps

### DIFF
--- a/template.Dockerfile
+++ b/template.Dockerfile
@@ -13,34 +13,17 @@ ENV PYTHONFAULTHANDLER=1 \
   VIRTUALENV_PIP=21.2.1 \
   MYPY_VERSION=0.930
 
-
 RUN pip install "poetry==$POETRY_VERSION" "mypy==$MYPY_VERSION"
-# pylintrc, coveragerc, poetry.lock and pyproject.toml shall not change very
-# often, so it is a good idea to add them as soon as possible
-COPY .coveragerc mypy.ini .flake8  ./
-COPY poetry.lock pyproject.toml ./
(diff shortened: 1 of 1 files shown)


template.Dockerfile
```diff
@@ -13,35 +13,18 @@ ENV PYTHONFAULTHANDLER=1 \
...
 
 # Copy the project to the system
 COPY iso15118/ iso15118/
 
-# Run the tests and linting
-COPY tests/ tests/
-RUN poetry run black --check --diff --line-length=88 iso15118 tests
-RUN poetry run flake8 --config .flake8 iso15118 tests
-# RUN poetry run mypy --config-file mypy.ini iso15118 tests
-RUN poetry run pytest -vv --cov-config .coveragerc --cov-report term-missing  --durations=3 --cov=.
```
tropxy, line 42: why did you remove this?

tropxy (changes requested):

template.Dockerfile
mdwcrft: I'm not sure it's good practice to run tests as part of a docker image build, the flow mentioned here is more standard: https://stackoverflow.com/questions/67088512/should-i-run-tests-during-the-docker-build

tropxy: ok, I see the point, but if we remove it from the dockerfile then we need to provide an alternative to run the tests, otherwise there wont be running tests

mdwcrft: Done in #147

tropxy:
can you rebase?

PR #111, opened 2022-08-18
feat: update python to 3.10 in pyproject toml

mdwcrft:
Updates pyproject.toml and mypy to 3.10, Dockerfile and Github Actions already use 3.10


tropxy:
rebase and then we can approve it

tropxy (approved):

PR #115, opened 2022-08-24
feat: Enable TLS 1.3 with mutual auth (AB:2378)

shalinnijel2:
Summary of changes:
- Updated create_certs.sh to generate oemCertChain.pem
- Added a new env variable ENABLE_TLS_1_3 - False by default
- Updated get_ssl_context() for server to check client cert chain and client to offer cert chain if TLS 1.3 is enabled 
- Updated README.
- Updated .env.dev.local and .env.dev.docker to include ENABLE_TLS_1_3 variable.


iso15118/shared/security.py
```diff
@@ -125,8 +126,14 @@ def get_ssl_context(server_side: bool) -> Optional[SSLContext]:
          Need to figure out a way to securely store those certs and keys
          as well as read the password.
     """
-    # TODO In ISO 15118-20, we use TLS 1.3. Need to adapt that later
-    ssl_context = SSLContext(protocol=PROTOCOL_TLSv1_2)
+
+    if ENABLE_TLS_1_3:
+        ssl_context = ssl.create_default_context(
+            purpose=Purpose.CLIENT_AUTH if server_side else Purpose.SERVER_AUTH,
+            cafile=CertPath.OEM_ROOT_PEM if server_side else CertPath.V2G_ROOT_PEM,
```
tropxy, line 133: shouldnt it be:

```python
cafile=CertPath.V2G_ROOT_PEM if server_side else CertPath.OEM_ROOT_PEM,
```

shalinnijel2: It looks like caFile points to the certificate required to verify the peer's chain ...
https://docs.python.org/3/library/ssl.html#ssl.create_default_context

iso15118/shared/security.py
```diff
@@ -125,8 +126,14 @@ def get_ssl_context(server_side: bool) -> Optional[SSLContext]:
...
+        )
+    else:
+        ssl_context = SSLContext(protocol=ssl.PROTOCOL_TLSv1_2)
```
tropxy, line 136:
```suggestion
        ssl_context = SSLContext(protocol=ssl.PROTOCOL_TLS)
```

iso15118/shared/security.py
tropxy: as discussed, the create_default_context already does the same settings as later done in the code; instead we can do the following:

```python
if ENABLE_TLS_1_3:
    ssl_context = SSLContext(protocol=ssl.PROTOCOL_TLS)
else:
    ssl_context = SSLContext(protocol=ssl.PROTOCOL_TLSv1_2)
```

shalinnijel2: Sure, done.

iso15118/shared/security.py
```diff
@@ -125,8 +126,16 @@ def get_ssl_context(server_side: bool) -> Optional[SSLContext]:
          Need to figure out a way to securely store those certs and keys
          as well as read the password.
     """
-    # TODO In ISO 15118-20, we use TLS 1.3. Need to adapt that later
-    ssl_context = SSLContext(protocol=PROTOCOL_TLSv1_2)
+
+    if ENABLE_TLS_1_3:
+        ssl_context = ssl.SSLContext()
```
tropxy, line 131: creating the context without specifying the protocol is deprecated. select ssl.PROTOCOL_TLS instead

tropxy (changes requested):

iso15118/shared/security.py
shalinnijel2: Oops...changed now...

tropxy (approved):

PR #120, opened 2022-08-31
Fix/ocsp extraction error raised

### LINKED ISSUE #96: Handle missing responder_url in a better way
Currently, if the responder_url cannot be obtained from a certificate, we just put "https://www.example.com".

This is probably not reasonable since it just pushes the error elsewhere.

Handle this case in a better way, if possible -- it may be that we can also fall back here on the CSMS retrieving the root certificate, or trying a URL of another certificate in the chain (are these sometimes left out inadvertently in practice?)

tropxy:
Added the OCSP field to the config when generating new certs
Added a dummy certificate with no OCSP data for testing purposes
Refactored the logic to extract the hash data to be more legible
No OCSP data means there is no need to build hash data for the certificate, as there is no responder URL to perform the call...

Fix #96


iso15118/shared/security.py
```diff
@@ -1235,11 +1235,8 @@ def derive_certificate_hash_data(
 
     try:
         responder_url = get_ocsp_url_for_certificate(certificate)
-    except (ExtensionNotFound, OCSPServerNotFoundError):
-        # TODO GitHub#96: This may just result in failure down the road.
-        # Should we let this fail on these exceptions, or is there
-        # another way to try to get a responder_url?
-        responder_url = "https://www.example.com/"
+    except (ExtensionNotFound, OCSPServerNotFoundError) as e:
+        raise e
```
danielgordon-switch-ev, line 1239: That works -- you can close #96 after this is merged.

iso15118/shared/security.py
```diff
@@ -1323,8 +1320,8 @@ def get_ocsp_url_for_certificate(certificate: Certificate) -> str:
 
 
 def all_certificates_from_chain(
-    certificate_chain: CertificateChainV2, root_cert: Optional[Certificate]
-) -> List[Certificate]:
+    certificate_chain: CertificateChainV2, root_cert: Optional[bytes]
+) -> List[bytes]:
```
danielgordon-switch-ev, line 1324: What's the reason to change away from the Certificate type here and elsewhere?

iso15118/shared/security.py
```diff
@@ -1376,21 +1373,29 @@ def get_certificate_hash_data(
...
+            if idx < len(all_certificates) - 1:
+                hash_data.append(
+                    derive_certificate_hash_data(certificate, all_certificates[idx + 1])
+                )
+            else:
+                # the last entry of the list contains the root_cert which
+                # is a self-signed certificate
+                hash_data.append(derive_certificate_hash_data(root_cert, root_cert))
+        except (ExtensionNotFound, OCSPServerNotFoundError):
+            # if we cant extract the OCSP data, then we ignore it and move on
```
danielgordon-switch-ev, line 1392: Does this help anything?  If one of the sub-CA certificates is missing its responder URL, this will assemble the rest of the chain, but it won't be a complete one.  Wouldn't it just fail anyway later on?

iso15118/shared/pki/configs/contractLeafCert.cnf
```diff
@@ -12,4 +12,4 @@ domainComponent			= MO
 basicConstraints		= critical,CA:false
 keyUsage				= critical,digitalSignature,nonRepudiation,keyEncipherment,keyAgreement
 subjectKeyIdentifier	= hash
-
+authorityInfoAccess = OCSP;URI:http://ca.myown.com/, caIssuers;URI:http://ca.myown.com/Intermediate-CA.cer
```
danielgordon-switch-ev, line 15: Like example.com, this seems not to be a real site -- is there a benefit to replacing the canonical fake site with one where this fact is less obvious?

danielgordon-switch-ev (changes requested):
A few questions, but generally this looks OK.

iso15118/shared/security.py
tropxy: the data type was wrong... the confusion was that there are two Certificate datatypes:
1. https://github.com/SwitchEV/iso15118/blob/master/iso15118/shared/messages/iso15118_2/datatypes.py#L53
2. https://github.com/SwitchEV/iso15118/blob/master/iso15118/shared/security.py#L34

The ISO15118 data type is of the type of the 1. which is bytes and not of the type from the Cryptography library, which was what was being used in the security.py file

iso15118/shared/security.py
tropxy: I was wondering what should be the best... we may have enough data to request the OCSP result for some of the certificates but not all of them, so shall we do it or shall we just completely drop this action and do not do any OCSP call?

iso15118/shared/pki/configs/contractLeafCert.cnf
tropxy: no, not really, was just the example I extracted... I can set it back to example

iso15118/shared/security.py
tropxy: I reflected about it and you are right, there is no point on returning an incomplete hash data, it is all or nothing

iso15118/shared/pki/configs/contractLeafCert.cnf
danielgordon-switch-ev: Probably better -- `ca.myown.com` isn't currently registered, but `myown.com` redirects to oprah.com, the site of Oprah Winfrey, an American talk show host.

iso15118/shared/security.py
danielgordon-switch-ev: I see -- is it worth adding a ticket to rename the ISO 15118 type to, say, `CertificateBytes` (or doing this as a follow-up PR if it's simple)?

iso15118/shared/security.py
```diff
@@ -1305,7 +1306,7 @@ def get_ocsp_url_for_certificate(certificate: Certificate) -> str:
             ExtensionOID.AUTHORITY_INFORMATION_ACCESS
         ).value
     except ExtensionNotFound:
-        logger.exception("Authority Information Access extension not found.")
+        logger.debug("Authority Information Access extension not found.")
```
danielgordon-switch-ev, line 1309: I don't understand your comment in the commit message that having this is optional and not an exception -- since we are doing OCSP validation, we need to know the address of the OCSP responder.

However, this may be worth doing anyway since using `logger.exception` here will just cause the traceback to be logged again -- that would be redundant with logging this if the error is uncaught, or is caught later on and logged.  Is `debug` the right level, though?

iso15118/shared/security.py
```diff
@@ -1376,21 +1377,30 @@ def get_certificate_hash_data(
...
+                    derive_certificate_hash_data(certificate, all_certificates[idx + 1])
+                )
+            else:
+                # the last entry of the list contains the root_cert, which
+                # is a self-signed certificate
+                hash_data.append(derive_certificate_hash_data(root_cert, root_cert))
+    except (ExtensionNotFound, OCSPServerNotFoundError):
+        # if we cant extract the OCSP from one of the certificates,
+        # then there is no point of building the hash data
+        return
```
danielgordon-switch-ev, line 1398: Minor, but `return None` might be better for clarity.

iso15118/shared/security.py
tropxy: yes, let's do it in another PR

danielgordon-switch-ev (approved):

iso15118/shared/security.py
tropxy: we discussed this during a face time call

PR #135, opened 2022-09-16
genchallange check has been added for Authorization

ikaratass:
fix [AB#2850](https://dev.azure.com/switch-ev/09ab882d-ca4b-4b5b-b5b8-80603c12e5b2/_workitems/edit/2850)

According to spec 15118-2  [V2G2-475], the GenChallenge in AuthorizationReq should be the same in PaymentDetailsRes. Otherwise, we need to return FAILED_ChallengeInvalid.

[V2G2-475] The message 'AuthorizationRes' shall contain the ResponseCode
'FAILED_ChallengeInvalid' if the challenge response contained in the
 AuthorizationReq message in attribute GenChallenge is not valid versus
the provided GenChallenge in PaymentDetailsRes.

### DIFF
--- a/iso15118/secc/states/iso15118_2_states.py
+++ b/iso15118/secc/states/iso15118_2_states.py
@@ -1021,14 +1021,13 @@ class Authorization(StateSECC):
         # 'FAILED_ChallengeInvalid' if the challenge response contained in the
         # AuthorizationReq message in attribute GenChallenge is not valid versus
         # the provided GenChallenge in PaymentDetailsRes.
-        if authorization_req.gen_challenge:
-            if authorization_req.gen_challenge != self.comm_session.gen_challenge:
-                self.stop_state_machine(
-                    "[V2G2-475] GenChallenge is not the same in PaymentDetailsRes",
-                    message,
-                    ResponseCode.FAILED_CHALLENGE_INVALID,
-                )
-                return
+        if authorization_req.gen_challenge != self.comm_session.gen_challenge:
+            self.stop_state_machine(
+                "[V2G2-475] GenChallenge is not the same in PaymentDetailsRes",
+                message,
+                ResponseCode.FAILED_CHALLENGE_INVALID,
+            )
+            return
 
         if self.comm_session.selected_auth_option == AuthEnum.PNC_V2:
             if not self.comm_session.contract_cert_chain:
--- a/iso15118/secc/states/iso15118_2_states.py
+++ b/iso15118/secc/states/iso15118_2_states.py
@@ -914,10 +914,10 @@ class PaymentDetails(StateSECC):
                 AuthorizationStatus.ACCEPTED,
                 AuthorizationStatus.ONGOING,
             ]:
-
+                self.comm_session.gen_challenge = get_random_bytes(16)
                 payment_details_res = PaymentDetailsRes(
                     response_code=ResponseCode.OK,
-                    gen_challenge=get_random_bytes(16),
+                    gen_challenge=self.comm_session.gen_challenge,
                     evse_timestamp=time.time(),
                 )
 
@@ -1016,6 +1016,19 @@ class Authorization(StateSECC):
             return
 
         authorization_req: AuthorizationReq = msg.body.authorization_req
+
+        # [V2G2-475] The message 'AuthorizationRes' shall contain the ResponseCode
+        # 'FAILED_ChallengeInvalid' if the challenge response contained in the
+        # AuthorizationReq message in attribute GenChallenge is not valid versus
+        # the provided GenChallenge in PaymentDetailsRes.
+        if authorization_req.gen_challenge:
+            if authorization_req.gen_challenge != self.comm_session.gen_challenge:
+                self.stop_state_machine(
+                    "[V2G2-475] GenChallenge is not the same in PaymentDetailsRes",
+                    message,
+                    ResponseCode.FAILED_CHALLENGE_INVALID,
+                )
+                return
 
         if self.comm_session.selected_auth_option == AuthEnum.PNC_V2:
             if not self.comm_session.contract_cert_chain:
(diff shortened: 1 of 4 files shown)


iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1013,6 +1013,19 @@ async def process_message(
...
         authorization_req: AuthorizationReq = msg.body.authorization_req
 
+        # [V2G2-475] The message 'AuthorizationRes' shall contain the ResponseCode
+        # 'FAILED_ChallengeInvalid' if the challenge response contained in the
+        # AuthorizationReq message in attribute GenChallenge is not valid versus
+        # the provided GenChallenge in PaymentDetailsRes.
+        if authorization_req.gen_challenge:
+            if authorization_req.gen_challenge != self.comm_session.gen_challenge:
+                self.stop_state_machine(
+                    "GenChallenge is not the same in AuthorizationRes",
```
shalinnijel2, line 1023: I believe you meant "GenChallenge not the same as what was sent in PaymentDetailsRes", right - not AuthorizationRes.

shalinnijel2 (changes requested):

iso15118/secc/states/iso15118_2_states.py
ikaratass: it is changed as PaymentDetailsRes. Thanks...

tests/secc/states/test_iso15118_2_states.py
```diff
@@ -193,6 +194,23 @@ async def test_authorization_next_state_on_authorization_request(
             == expected_evse_processing
         )
 
+    async def test_authorization_req_gen_challange(self):
+        self.comm_session.writer = Mock()
```
tropxy, line 198: the test name should include the suffix invalid. Also I think you should create a test where the outcome is valid, if there's none yet

tropxy (changes requested):

tests/secc/states/test_iso15118_2_states.py
ikaratass: Done

shalinnijel2 (dismissed):

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1013,6 +1013,19 @@ async def process_message(
...
         authorization_req: AuthorizationReq = msg.body.authorization_req
 
+        # [V2G2-475] The message 'AuthorizationRes' shall contain the ResponseCode
+        # 'FAILED_ChallengeInvalid' if the challenge response contained in the
+        # AuthorizationReq message in attribute GenChallenge is not valid versus
+        # the provided GenChallenge in PaymentDetailsRes.
+        if authorization_req.gen_challenge:
+            if authorization_req.gen_challenge != self.comm_session.gen_challenge:
+                self.stop_state_machine(
+                    "GenChallenge is not the same in PaymentDetailsRes",
```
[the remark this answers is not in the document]
ikaratass: [V2G2-475] is added to in message.

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1021,6 +1021,19 @@ async def process_message(
 
         authorization_req: AuthorizationReq = msg.body.authorization_req
 
+        # [V2G2-475] The message 'AuthorizationRes' shall contain the ResponseCode
+        # 'FAILED_ChallengeInvalid' if the challenge response contained in the
+        # AuthorizationReq message in attribute GenChallenge is not valid versus
+        # the provided GenChallenge in PaymentDetailsRes.
+        if authorization_req.gen_challenge:
```
tropxy, line 1028: if the field gen_challenge is addressable but can be None, then do we need this if clause?
cant we just have the one after?

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
ikaratass: i see, gen_challenge is optional, if we remove this statement, we can face an issue when ev doesnt send gen_challenge value. if ev doesnt send it, the value will be None, and automatically we return as FAILED_CHALLENGE_INVALID, that is not a valid case. Am i wrong?

tropxy: you are not checking if the gen_challenge attribute exists in the authorization_req data structure. Instead, you are already assuming that the attribute exists but that its value can be None or not: this may be ok, depends how the pydantic model constructs the data. Is it the case that if the gen_challenge is None we will still be able to access the `gen_challenge`attribute? if so, then we dont need that first if clause, because in the one after we do the check as well

shalinnijel2: There is one use case being missed because of the if condition above - the case where GenChallenge is not send at all in AuthorizationReq in PnC.

I believe you added the check as the field is optional in EIM. It would continue to work in EIM if the check is removed anyway because of the reason @Andre mentioned above - but might throw an error in case a GenChallenge is accidentally sent in EIM AuthorizationReq. It may be better to ignore the field check in EIM completely?

```
if self.comm_session.selected_auth_option is AuthEnum.PNC_V2:
```

ikaratass: @shalinnijel2 should not we terminate the session with FAILED_CHALLENGE_INVALID in this case as well. I think, we might not need this check...

shalinnijel2: Yes, true  - letting that pass through would also be a violation of the spec - in which case simply removing the if condition as discussed above would suffice.

tests/secc/states/test_messages.py
```diff
@@ -118,11 +118,15 @@ def get_dummy_v2g_message_session_stop_req():
...
+    id: Optional[str] = None, gen_challenge: Optional[bytes] = None
+):
     # The AuthorizationReq is empty, unless it is following a PaymentDetailsRes
     # message, in which case it must send back the generated challenge.
-    authorization_req = AuthorizationReq()
-
+    if gen_challenge:
+        authorization_req = AuthorizationReq(id=id, gen_challenge=gen_challenge)
+    else:
+        authorization_req = AuthorizationReq()
```
tropxy, line 129: if you are not generating here the challenge and just injecting, there is no point on having the if clause.
I would just pass along the injection, but is a minor thing

tropxy (approved):

PR #139, opened 2022-09-30
Feat/improve logging/ab#2898

santiagosalamandri:
This PR modifies Log levels:

- INFO level for TCP/TLS messages
- DEGUG level for 'waiting for up to ... sec' message
- INFO level for "chosen protocol" message
- INFO level for details on Request/Response messages

It also adds logs for project settings (env. variables) at start-up. It includes shared and EVCC/SECC-specific settings.

### DIFF
--- a/iso15118/shared/settings.py
+++ b/iso15118/shared/settings.py
@@ -23,4 +23,6 @@ V20_EVSE_SERVICES_CONFIG = env.str(
 )
 
 ENABLE_TLS_1_3 = env.bool("ENABLE_TLS_1_3", default=False)
+
+shared_settings=env.dump()
 env.seal()  # raise all errors at once, if any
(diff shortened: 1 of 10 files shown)


iso15118/shared/settings.py
```diff
@@ -23,4 +23,6 @@
 )
 
 ENABLE_TLS_1_3 = env.bool("ENABLE_TLS_1_3", default=False)
+
+shared_settings = env.dump()
```
shalinnijel2, line 27: One minor suggestion - would be good to call env.dump() after env.seal()?

shalinnijel2 (changes requested):

shalinnijel2 (approved):

iso15118/shared/settings.py
santiagosalamandri: you're right. Done.

PR #146, opened 2022-10-12
get from the evse controller the ac evse status

tropxy:
ac evse status was static during the charging loop, so I modified the code to make it dynamic

### DIFF
--- a/tests/secc/states/test_iso15118_2_states.py
+++ b/tests/secc/states/test_iso15118_2_states.py
@@ -7,6 +7,7 @@ from iso15118.secc import Config
 from iso15118.secc.states.iso15118_2_states import (
     Authorization,
     ChargeParameterDiscovery,
+    ChargingStatus,
     CurrentDemand,
     PaymentDetails,
     PowerDelivery,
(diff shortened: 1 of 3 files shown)


shalinnijel2 (approved):

tests/secc/states/test_iso15118_2_states.py
```diff
@@ -383,3 +386,35 @@ async def test_service_discovery_req_unexpected_state(self):
...
+        charging_status = ChargingStatus(self.comm_session)
+        self.comm_session.writer = Mock()
+        self.comm_session.writer.get_extra_info = Mock()
+        self.comm_session.selected_schedule = 1
+
+        async def get_ac_evse_status_patch():
+            return ACEVSEStatus(
+                notification_max_delay=0,
+                evse_notification=EVSENotification.NONE,
+                rcd=True,
```
shalinnijel2, line 414: Ok - I see you changed the rcd value for this response to be different from the simulator...

shalinnijel2 (approved):

Issue #93, opened 2022-08-04
Use mypy


danielgordon-switch-ev:
In our experience, mypy is very useful to defend against type-related bugs in Python programs.  We are using it elsewhere, and it would be a good addition to the iso15118 library.

mdwcrft:
Mypy was enabled in #294

PR #147, opened 2022-10-17
feat: run code qual and tests in gha

mdwcrft:
- Add mypy as a dev dependency
- Update makefile to run code qual via poetry
- Run code qual and tests as part of GHA
- MyPy isn't run yet as it will fail, this should be addressed as part of the issue #93


tropxy (approved):
LGTM just needs a rebase

PR #148, opened 2022-10-17
feat/service_status/ab#2950

santiagosalamandri:
This PR allows setting and logging service status based on the servers' statuses:
  
- The condition for the service to be ready is that all servers are up and running. This condition is checked every 100 mS.
- There is a 10-second timeout to notify that not all servers have been started.

### DIFF
--- a/iso15118/secc/comm_session_handler.py
+++ b/iso15118/secc/comm_session_handler.py
@@ -16,7 +16,10 @@ import socket
 from asyncio.streams import StreamReader, StreamWriter
 from typing import Dict, List, Optional, Tuple, Union
 
-from iso15118.secc.controller.interface import EVSEControllerInterface
+from iso15118.secc.controller.interface import (
+    EVSEControllerInterface,
+    EVSEServiceStatus,
+)
 from iso15118.secc.failed_responses import (
     init_failed_responses_din_spec_70121,
     init_failed_responses_iso_v2,
@@ -168,6 +171,9 @@ class CommunicationSessionHandler:
         self.config = config
         self.evse_controller = evse_controller
 
+        # List of server status events
+        self.status_event_list: List[asyncio.Event] = []
+
         # Set the selected EXI codec implementation
         EXI().set_exi_codec(codec)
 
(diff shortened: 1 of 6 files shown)


iso15118/secc/transport/udp_server.py
```diff
@@ -99,6 +100,7 @@ async def start(self):
             f"{SDP_MULTICAST_GROUP}%{self.iface} "
             f"and port {SDP_SERVER_PORT}"
         )
+        self.ready_event.set()
```
shalinnijel2, line 103: I'm not sure if there is much value in making read_event a member variable as the only thing time it is used for is to indicate when the udp server is setup? Wouldn't it suffice to pass it through the start() method?

iso15118/secc/transport/tcp_server.py
```diff
@@ -112,6 +114,11 @@ async def server_factory(self, tls: bool) -> None:
             f"port {port}"
         )
 
+        if tls:
+            self.tls_ready_event.set()
+        else:
+            self.tcp_ready_event.set()
+
```
shalinnijel2, line 121: Same comment regarding udp server as mentioned below. Wouldn't it suffice if we pass in the ready_event Event through the start_tls() or start_no_tls() method - as the state is not required after this - so making it a member variable would not be required. What do you think?

iso15118/secc/controller/interface.py
```diff
@@ -72,6 +73,14 @@ class EVDataContext:
     soc: Optional[int] = None  # 0-100
 
 
+class EVSEServiceStatus(str, Enum):
```
shalinnijel2, line 76: As the intention of the enum would be to indicate the status of the entire 15118 service and not just the one EVSE (even though that is what it is being used for at the moment but would change soon), I think it might be better to reflect that - calling it ServiceStatus should do?

shalinnijel2 (changes requested):

iso15118/secc/comm_session_handler.py
```diff
@@ -189,21 +195,47 @@ async def __init__.
...
+        for event in self.status_event_list:
+            if event.is_set() is False:
+                result = False
+                break
+        return result
+
+    async def check_ready_status(self) -> None:
+        # Wait until all flags are set
+        while self.check_events() is False:
+            await asyncio.sleep(0.1)
```
shalinnijel2, line 229: I think 100 ms maybe too liberal for this - as servers could potentially be ready in just a few ms - modify this to be 10ms maybe?

iso15118/secc/transport/udp_server.py
santiagosalamandri: Done.

iso15118/secc/transport/tcp_server.py
santiagosalamandri: I thought it could be used in the future to notify when a server failed. It's not needed for now.

iso15118/secc/controller/interface.py
santiagosalamandri: Done

iso15118/secc/comm_session_handler.py
santiagosalamandri: Sure.

shalinnijel2 (approved):

PR #150, opened 2022-10-19
Fix tc secc ac vtb power delivery 010

ikaratass:
fix [AB#2853](https://dev.azure.com/switch-ev/09ab882d-ca4b-4b5b-b5b8-80603c12e5b2/_workitems/edit/2853) 

[V2G2-225] The SECC shall send the negative ResponseCode
FAILED_ChargingProfileInvalid in the PowerDelivery response message if the EVCC sends a ChargingProfile which
is not adhering to the PMax values of all PMaxScheduleEntry elements according
to the chosen SAScheduleTuple element in the last ChargeParameterDiscoveryRes
message sent by the SECC.

### DIFF
--- a/iso15118/secc/states/iso15118_2_states.py
+++ b/iso15118/secc/states/iso15118_2_states.py
@@ -1472,7 +1472,21 @@ class PowerDelivery(StateSECC):
         #     )
         #     return
 
-        # TODO We should also do a more detailed check of the charging profile
+        # [V2G2-225] The SECC shall send the negative ResponseCode
+        # FAILED_ChargingProfileInvalid in
+        # the PowerDelivery response message if the EVCC sends a ChargingProfile which
+        # is not adhering to the PMax values of all PMaxScheduleEntry elements according
+        # to the chosen SAScheduleTuple element in the last ChargeParameterDiscoveryRes
+        # message sent by the SECC.
+        if power_delivery_req.charging_profile:
+            if not self._is_charging_profile_valid(power_delivery_req):
+                self.stop_state_machine(
+                    "[V2G2-225] ChargingProfile is not adhering to the Pmax values in "
+                    "ChargeParameterDiscoveryRes",
+                    message,
+                    ResponseCode.FAILED_CHARGING_PROFILE_INVALID,
+                )
+                return
 
         logger.debug(f"ChargeProgress set to {power_delivery_req.charge_progress}")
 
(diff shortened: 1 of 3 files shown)


iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1458,6 +1458,33 @@ async def process_message(
...
+        # to the chosen SAScheduleTuple element in the last ChargeParameterDiscoveryRes
+        # message sent by the SECC.
+        if power_delivery_req.charging_profile:
+            for profile_entry in power_delivery_req.charging_profile.profile_entries:
+                for schedule in self.comm_session.offered_schedules:
+                    for p_max_profile_entry in schedule.p_max_schedule.schedule_entries:
+                        if (
+                            p_max_profile_entry.time_interval.start
+                            == profile_entry.start
+                        ):
```
shalinnijel2, line 1474: I think the check should be made against the chosen schedule in the offered schedule.
The following case is an example use case where the above check could pass but should fail:
Offered chosen schedule
(0s)<---5KW--->(10s)<---10KW<---(60s)--->End

Returned profile
(0s)<--4KW-->(2s)<--6KW-->(3s)<--10KW-->(60s)--->End

So it is necessary to pick every interval returned in the charging profile and make sure it stays below the limits of the interval(s) that entry might correspond to.

tests/secc/states/test_messages.py
```diff
@@ -201,6 +203,34 @@ def get_dummy_v2g_message_power_delivery_req_charge_start():
     )
 
 
+def get_v2g_message_power_delivery_req_charge_start_invalid_charging_profile():
+    charging_profile = ChargingProfile(
```
shalinnijel2, line 207: It would be good to add two tests that covers the above mentioned use case as well - which is return a charging profile that doesn't fall on the same boundary as the chosen schedule - but is in line with the limits and one that is not.

shalinnijel2 (changes requested):

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1458,6 +1458,33 @@ async def process_message(
...
+
+                                self.stop_state_machine(
+                                    f"[V2G2-225] {profile_entry.max_power.value} is "
+                                    f"greater than {p_max_profile_entry.p_max.value}",
+                                    message,
+                                    ResponseCode.FAILED_CHARGING_PROFILE_INVALID,
+                                )
+                                return
+
         # TODO We should also do a more detailed check of the charging profile
```
shalinnijel2, line 1488: This line could be removed as well.

iso15118/secc/states/iso15118_2_states.py
tropxy: I think the solution is something like this
```python
for schedule in self.comm_session.offered_schedules: 
    if schedule.sa_schedule_tuple_id == power_delivery_req.sa_schedule_tuple_id:
        sa_schedule_pmax = max([profile_entry.p_max.value * 10**profile_entry.p_max.multiplier for profile_entry in schedule.p_max_schedule.schedule_entries])
        ev_profile_pmax = max([profile_entry.max_power.value * 10**profile_entry.max_power.multiplier for profile_entry in power_delivery_req.charging_profile.profile_entries])
        if ev_profile_pmax > sa_schedule_pmax:
            self.stop_state_machine(....)
            return
```

I guess the code to find the max of each set could be optimized, maybe using `map` or something,,,

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
ikaratass: @tropxy should not we check each profile entry values according to time, and it should be lower than what offered in offered_schedules. if we only check the maximum value of given schedule, it would be invalid for the case below:

Offered chosen schedule
(0s)<---5KW--->(10s)<---10KW<---(60s)--->End

Returned profile
(0s)<--4KW-->(2s)<--6KW-->(3s)<--10KW-->(60s)--->End 

in this case, for the value 6KW should not exceed 5KW. This is what I understood.


do you think if this is a realistic case?

ikaratass: I have added a logic to cover the case above, but still, I am not happy with it...

tropxy: yeah, looking at this again, you have a point. I believe you are right, we have to go over the different slots of time of the chosen schedule (sa_schedule) and see if the power consumed by the EV (ev_profile) during that slot does not exceed the one set by the sa_schedule.

Thus, the logic I provided before is incorrect...

tropxy: This shall be the logic then, I believe....
```python
for schedule in self.comm_session.offered_schedules:
    if schedule.sa_schedule_tuple_id == power_delivery_req.sa_schedule_tuple_id:
        for sa_profile_entry in schedule.p_max_schedule.schedule_entries:
            for ev_profile_entry in power_delivery_req.charging_profile.profile_entries:
                if ev_profile_entry.start <= sa_profile_entry.time_interval.start:
                    sa_pmax = sa_profile_entry.p_max.value * 10 ** sa_profile_entry.p_max.multiplier
                    ev_pmax = ev_profile_entry.max_power.value * 10 ** ev_profile_entry.max_power.multiplier
                    if ev_pmax > sa_pmax:
                        self.stop_state_machine(...)
                        return
                else:
                    break

````

Comment (tropxy) [reply iso15118/secc/states/iso15118_2_states.py]: The picture of a schedule in Marcs book, helps on visualising what you need to do:
Bildschirmfoto 2022-10-24 um 19 40 29

Comment (ikaratass) [reply iso15118/secc/states/iso15118_2_states.py]: Done

File: iso15118/secc/states/iso15118_2_states.py [hunk 4]
```diff
@@ -1458,7 +1458,67 @@ async def process_message(
...
+                            try:
+                                # By getting the next entry/slot, we can know when
+                                # the current entry ends
+                                next_sa_profile_entry = schedule_entries[idx + 1]
+                                sa_profile_entry_end = (
+                                    next_sa_profile_entry.time_interval.start
+                                )
+                            except IndexError:
+                                # As the index is out of rage, it signals this is
+                                # the final entry of the sa profile the last entry
```
tropxy, line 1497:
```suggestion
# the final entry of the sa profile. The last entry
```

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1458,7 +1458,67 @@ async def process_message(
...
+                                # the end of the period summing the relative start
+                                # of the entry with the duration of it
+                                sa_profile_entry_end = (
+                                    sa_profile_entry_start
+                                    + sa_profile_entry.time_interval.duration
+                                )
+                            if (
+                                sa_profile_entry_start
+                                <= ev_profile_entry.start
+                                <= sa_profile_entry_end
```
tropxy, line 1508: this doesnt look readable...as it seems the problem is with the line break, we have three options:
1. either ignore the flake error by adding #noqa to the end and keeping the line as
`if sa_profile_entry_start <= ev_profile_entry.start <= sa_profile_entry_end:  #noqa`

2. or doing:
```python
if (
ev_profile_entry.start  >= sa_profile_entry_start and 
ev_profile_entry.start <= sa_profile_entry_end
):
```

3. This one is probably the best. As this part of the code is big and has too many indents (spaces), the best is to put the code in a private method within the class, like done here:
https://github.com/SwitchEV/iso15118/blob/0ee9d8ba853260d7ec38f8ff9b42339c3affa54b/iso15118/secc/states/iso15118_2_states.py#L1631

dont forget to prefix the "_" before the name of the private method

tropxy (changes requested):

iso15118/secc/states/iso15118_2_states.py
ikaratass: Even i add the` #noqa` end of the line, black is still not happy. 
image

ikaratass: fmt:on/off fixed this problem.

iso15118/secc/states/iso15118_2_states.py
```diff
@@ -1597,6 +1611,50 @@ async def check_state():
...
+                            sa_profile_entry_end = (
+                                sa_profile_entry_start
+                                + sa_profile_entry.time_interval.duration
+                            )
+                        # fmt: off
+                        if sa_profile_entry_start <= ev_profile_entry.start <= sa_profile_entry_end:  # noqa
+                            ev_entry_pmax = (ev_profile_entry.max_power.value
+                                             * 10 ** ev_profile_entry.max_power.multiplier)  # noqa
+                            if ev_entry_pmax > sa_entry_pmax:
+                                return False
```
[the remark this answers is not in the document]
tropxy: So, I went over this and realized that that was not the problem. the code was breaking the loop, but it needed some more logic to work as expected. I added it and also included a more elaborated test and looks good ;)

tropxy (approved):

iso15118/secc/states/iso15118_2_states.py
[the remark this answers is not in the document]
ikaratass: Yeah, I tested as well, the problem wasn't this, we should check all profile entries to find the correct range to be checked. it passed all unit tests. Thanks.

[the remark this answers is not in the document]
tropxy: yeah, we had to check all the entries with the logic before, but the more EV profile entries more time it would take as in the worst case scenario, if we have N EV profile entries and M SA schedule entries, we would do NxM checks all the time. And the rationale is that the EV entries we have already checked during one iteration, do not need to be reevaluated for the next one.

iso15118/secc/states/iso15118_2_states.py
ikaratass: Done

tests/secc/states/test_messages.py
ikaratass: done

shalinnijel2:
I think we have missed one use case where the ev profile spans over two sa schedule entries. In my test, the below given use case passed when it should have failed:
SA schedule: 

```
sa_schedule_tuple_id=1 p_max_schedule=PMaxSchedule(schedule_entries=[PMaxScheduleEntry(p_max=PVPMax(value=11000, multiplier=0, unit=<UnitSymbol.WATT: 'W'>), time_interval=RelativeTimeInterval(start=0, duration=None)), 
PMaxScheduleEntry(p_max=PVPMax(value=7000, multiplier=0, unit=<UnitSymbol.WATT: 'W'>), time_interval=RelativeTimeInterval(start=1800, duration=1800))]) 
```

EV Charging Profile:
```
[
ProfileEntryDetails(start=0, max_power=PVPMax(value=11000, multiplier=0, unit=<UnitSymbol.WATT: 'W'>), max_phases_in_use=3), 
ProfileEntryDetails(start=1200, max_power=PVPMax(value=11000, multiplier=0, unit=<UnitSymbol.WATT: 'W'>), max_phases_in_use=3), 
ProfileEntryDetails(start=1900, max_power=PVPMax(value=7000, multiplier=0, unit=<UnitSymbol.WATT: 'W'>), max_phases_in_use=3)]
```

Notice the second entry with Pmax 11000W spans over the 1800secs mark from which point only 7000W is available.

shalinnijel2 (approved):

PR #151, opened 2022-10-21
fix: use utcnow() to check certificate validity

rstanchak:
Symptom:
Using the certificates generated using ./create_certs.sh, 'make run-secc' with TLS and PNC enabled, would fail in the CertificateInstallationReq handling:

`ERROR    2022-10-21 14:37:20,294 - iso15118.shared.security (977): CertNotYetValidError: Signature verification failed while checking certificate chain`

Issue:
The cryptography Certificate object represents the validity period of the certificate in UTC.  https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate

However, security.py is comparing cert.not_valid_before to the local time, and as a result,  the certificate is reported invalid for those of us in zones earlier than GMT.

Fix:
Use datetime.utcnow() instead of datetime.now() to get the current time

### DIFF
--- a/iso15118/shared/security.py
+++ b/iso15118/shared/security.py
@@ -710,7 +710,7 @@ def check_validity(certs: List[Certificate]):
     Raises:
         CertNotYetValidError, CertExpiredError
     """
-    now = datetime.now()
+    now = datetime.utcnow()
     for cert in certs:
         if cert.not_valid_before > now:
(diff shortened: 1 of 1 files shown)


tropxy (approved):

tropxy:
thanks @rstanchak for your contribution!

PR #155, opened 2022-11-07
feat: interface added for pause and terminate

ikaratass:
* session_stop interface method  added for pause and terminate

### DIFF
--- a/iso15118/secc/controller/interface.py
+++ b/iso15118/secc/controller/interface.py
@@ -702,7 +702,7 @@ class EVSEControllerInterface(ABC):
         raise NotImplementedError
 
     @abstractmethod
-    async def session_stop(self, action: SessionStopAction) -> None:
+    async def update_data_link(self, action: SessionStopAction) -> None:
         """
         Called when EV requires termination or pausing of the charging session.
         Args:
--- a/iso15118/secc/controller/interface.py
+++ b/iso15118/secc/controller/interface.py
@@ -33,6 +33,7 @@ from iso15118.shared.messages.enums import (
     CpState,
     EnergyTransferModeEnum,
     Protocol,
+    SessionStopAction,
 )
 from iso15118.shared.messages.iso15118_2.datatypes import (
     ACEVSEChargeParameter,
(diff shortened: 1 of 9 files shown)


iso15118/shared/comm_session.py
```diff
@@ -391,8 +392,10 @@ async def stop(self, reason: str):
         logger.info(f"Reason: {reason}")
 
         await asyncio.sleep(2)
-        # TODO Signal data link layer to either terminate or pause the data
-        #      link connection
+        # Signal data link layer to either terminate or pause the data
+        # link connection
+        if hasattr(self.comm_session, "evse_controller"):
+            await self.comm_session.evse_controller.session_stop(terminate_or_pause)
```
shalinnijel2, line 398: To help test this feature (specifically pause), I'm thinking we could add a similar session_stop() on evcc side as well? If so, this would be modified to include an else condition for ev_controller as well.
```
if hasattr(self.comm_session, "evse_controller"):
  await self.comm_session.evse_controller.session_stop(terminate_or_pause)
elif hasattr(self.comm_session, "ev_controller"):
  await self.comm_session.ev_controller.session_stop(terminate_or_pause)
```
What do you think?

shalinnijel2 (changes requested):

iso15118/secc/controller/interface.py
```diff
@@ -699,3 +700,14 @@ async def get_15118_ev_certificate(
         - ISO 15118-20 and ISO 15118-2
         """
         raise NotImplementedError
+
+    @abstractmethod
+    async def session_stop(self, action: SessionStopAction) -> None:
```
shalinnijel2, line 705: I think it might be better to be more explicit about the intention of this method - say like update_data_link() - as session_stop() is likely to be misunderstood to be a method for stopping a 15118 session.

iso15118/shared/comm_session.py
ikaratass: As we talked, no need for ev_controller for now.

iso15118/secc/controller/interface.py
ikaratass: Done

shalinnijel2 (approved):

PR #159, opened 2022-11-10
Fix/set present

ikaratass:
* set_present_protocol_state method is awaitable.

### DIFF
--- a/iso15118/shared/comm_session.py
+++ b/iso15118/shared/comm_session.py
@@ -475,24 +475,19 @@ class V2GCommunicationSession(SessionStateMachine):
                 return
 
             try:
+                # This will create the values needed for the next state, such as
+                # next_state, next_v2gtp_message, next_message_payload_type etc.
+                await self.process_message(message)
                 if hasattr(self.comm_session, "evse_controller"):
                     await self.comm_session.evse_controller.set_present_protocol_state(
                         str(self.current_state)
                     )
-                # This will create the values needed for the next state, such as
-                # next_state, next_v2gtp_message, next_message_payload_type etc.
-                await self.process_message(message)
-
                 if self.current_state.next_v2gtp_msg:
                     # next_v2gtp_msg would not be set only if the next state is either
                     # Terminate or Pause on the EVCC side
                     await self.send(self.current_state.next_v2gtp_msg)
 
                 if self.current_state.next_state in (Terminate, Pause):
-                    if hasattr(self.comm_session, "evse_controller"):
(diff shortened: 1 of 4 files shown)


iso15118/shared/comm_session.py
```diff
@@ -475,6 +475,10 @@ async def rcv_loop(self, timeout: float):
                 return
 
             try:
+                if hasattr(self.comm_session, "evse_controller"):
+                    await self.comm_session.evse_controller.set_present_protocol_state(
+                        str(self.current_state)
+                    )
                 # This will create the values needed for the next state, such as
                 # next_state, next_v2gtp_message, next_message_payload_type etc.
                 await self.process_message(message)
```
shalinnijel2, line 484: Looks like it might suffice to call set_present_protocol_state once after the above self.process_message?
```
                if hasattr(self.comm_session, "evse_controller"):
                    await self.comm_session.evse_controller.set_present_protocol_state(
                        # noqa
                        str(self.current_state)
                    )
```

iso15118/shared/comm_session.py
```diff
@@ -475,6 +475,10 @@ async def rcv_loop(self, timeout: float):
                 return
 
             try:
+                if hasattr(self.comm_session, "evse_controller"):
+                    await self.comm_session.evse_controller.set_present_protocol_state(
+                        str(self.current_state)
+                    )
```
shalinnijel2, line 481: As discussed, this could be removed.

iso15118/shared/comm_session.py
```diff
@@ -485,6 +489,10 @@ async def rcv_loop(self, timeout: float):
                     await self.send(self.current_state.next_v2gtp_msg)
 
                 if self.current_state.next_state in (Terminate, Pause):
+                    if hasattr(self.comm_session, "evse_controller"):
+                        await self.comm_session.evse_controller.set_present_protocol_state(  # noqa
+                            str(self.current_state.next_state.__name__)
+                        )
```
shalinnijel2, line 495: This could be removed as discussed.

shalinnijel2 (changes requested):

iso15118/shared/comm_session.py
ikaratass: Done

iso15118/shared/comm_session.py
ikaratass: Done

iso15118/shared/comm_session.py
ikaratass: Done

shalinnijel2 (approved):

PR #161, opened 2022-11-11
Update udp socket to bind to specific interface

shalinnijel2:
Summary of changes:
- UDP socket now binds to the specific interface as defined in .env
- The interface to be used could be passed via start_secc_handler(). This helps remove reliance on env file when required.


shalinnijel2:
The way I tested this was by running two instances of iso15118 sessions on two different interfaces (2 secc-evcc pairs.)

PR #163, opened 2022-11-17
Feat/external auth

ikaratass:
it is removed Contract from the PaymentOptionsList when external authorization is done.

### DIFF
--- a/iso15118/secc/controller/interface.py
+++ b/iso15118/secc/controller/interface.py
@@ -200,10 +200,8 @@ class EVSEControllerInterface(ABC):
 
     async def is_external_authorization_done(self) -> bool:
         """
-        it returns true when an RFID card tapped before plugging in
-        and authorization is successful afterwards.
+        it returns true when an external authentication before plugging in.
         Relevant for:
-        - DIN SPEC 70121
         - ISO 15118-2
         - ISO 15118-20
         """
--- a/iso15118/secc/controller/interface.py
+++ b/iso15118/secc/controller/interface.py
@@ -198,6 +198,17 @@ class EVSEControllerInterface(ABC):
         """
         raise NotImplementedError
 
+    async def is_external_authorization_done(self) -> bool:
+        """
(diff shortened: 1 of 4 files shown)


iso15118/secc/controller/simulator.py
```diff
@@ -413,6 +413,10 @@ async def get_energy_service_list(self) -> ServiceList:
 
         return service_list
 
+    async def is_external_authorization_done(self) -> bool:
+        """Overrides EVSEControllerInterface.is_external_authorization_done()."""
+        return True
```
shalinnijel2, line 418: It would be better to leave the simulator at False as otherwise Contract wouldn't be offered by default in the simulator?

shalinnijel2, line 418: Wouldn't it be better to leave the simulator at False as otherwise Contract wouldn't be offered by default in the simulator?

shalinnijel2 (changes requested):

shalinnijel2:
Apart from the change in the simulator, the rest lgtm.

iso15118/secc/controller/simulator.py
ikaratass: Done

tropxy (changes requested):

iso15118/secc/controller/interface.py
```diff
@@ -198,6 +198,17 @@ async def get_energy_service_list(self) -> ServiceList:
...
         raise NotImplementedError
 
+    async def is_external_authorization_done(self) -> bool:
+        """
+        it returns true when an RFID card tapped before plugging in
+        and authorization is successful afterwards.
+        Relevant for:
+        - DIN SPEC 70121
+        - ISO 15118-2
+        - ISO 15118-20
```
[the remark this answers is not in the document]
ikaratass: Done

shalinnijel2 (approved):

tropxy (approved):

PR #164, opened 2022-11-21
Fix: UDP server bind issue after PR#161

shalinnijel2:
Summary of changes:
Use SO_BINDTODEVICE on linux to bind udp server to the specified interface. No changes to behaviour on Mac.

### DIFF
--- a/iso15118/secc/transport/udp_server.py
+++ b/iso15118/secc/transport/udp_server.py
@@ -58,8 +58,9 @@ class UDPServer(asyncio.DatagramProtocol):
         # Socket type (datagram, determines transport layer protocol UDP)
         sock = socket.socket(family=socket.AF_INET6, type=socket.SOCK_DGRAM)
 
-        # Allows address to be reused
-        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        # Block binding to this socket+interface combination from now.
+        # Ref: https://www.man7.org/linux/man-pages/man7/socket.7.html
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 0)
 
         # Bind the socket to the predefined port on specified interface for receiving
         # UDP packets (SDP requests). This is done differently on Mac and Linux.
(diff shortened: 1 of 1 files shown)


iso15118/secc/transport/udp_server.py
```diff
@@ -60,10 +61,24 @@ async def __init__.
...
+        # Reference:
+        # https://djangocas.dev/blog/linux/linux-SO_BINDTODEVICE-and-mac-IP_BOUND_IF-to-bind-socket-to-a-network-interface/ # noqa
+        # https://linux.die.net/man/7/socket
+        # https://stackoverflow.com/questions/20616029/os-x-equivalent-of-so-bindtodevice # noqa
+        if platform == "darwin":
+            full_ipv6_address = await get_link_local_full_addr(SDP_SERVER_PORT, iface)
+            sock.bind(full_ipv6_address)
+        else:
+            if not hasattr(socket, "SO_BINDTODEVICE"):
+                socket.SO_BINDTODEVICE = 25
```
tropxy, line 75: you can import IN:
```python
import IN

sock.setsockopt(socket.SOL_SOCKET, IN.SO_BINDTODEVICE, ...)
```

shalinnijel2: From the tests done on the board today, the hasattr check isn't required either. Have removed the check.

iso15118/secc/transport/udp_server.py
```diff
@@ -60,10 +61,22 @@ async def __init__.
         # Allows address to be reused
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
```
tropxy, line 62: we need to change this to 0, so that we block the case where we try to create a socket for the same interface. Also add a reference to the man page: https://www.man7.org/linux/man-pages/man7/socket.7.html

tropxy (changes requested):

iso15118/secc/transport/udp_server.py
shalinnijel2: Done.

tropxy (approved):

PR #169, opened 2022-11-28
Support 15118-20 DC BPT

shalinnijel2:
Summary of changes:
1. Added 15118-20 DC to supported protocols list
2. Fixed various ChargeLoop struct definitions.
3. Implemented the below states:
    1. DCCableCheck
    2. DCPreCharge
    3. DCWeldingDetection
    4. DCChargeLoop
4. Added tests


iso15118/secc/controller/interface.py
```diff
@@ -689,6 +729,54 @@ async def get_dc_bpt_charge_params_v20(
...
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def get_bpt_dynamic_dc_charge_loop_params(
+        self,
+    ) -> BPTDynamicDCChargeLoopRes:
+        """
+        Gets the parameters for the DCChargeLoopRes in the Dynamic control mode for
+        bi-directional power transfer (BPT)
```
tropxy, line 773: I dont think we need this level of segregation. We can get what we need with one call (named for example `get_evse_dc_limits` whose result shall encapsulate:
* EVSEMaxChargePower 
* EVSEMinChargePower
* EVSEMaxChargeCurrent
* EVSEMaxVoltage
* EVSEMinVoltage

If the EVSE does NOT support BPT, then the fields in the MQTT corresponding to the following ones can be set to 0. I would do that instead of allowing the values to be None, so that in case we are on a BPT session we always have values. 

* EVSEMaxDischargePower 
* EVSEMinDischargePower
* EVSEMaxDischargeCurrent


And actually, thinking about this, we should not request this values, neither the DepartureTime, neither the TargetSOC or MinSOC. All these values and the above ones shall be pushed, either by the manufacturer or by OCPP. Because, in the practical situation we may or may not have new targets and if it is the same why are we constantly sending a MQTT call and wasting time and cpu resources?

So, what we shall do is to request this call once: https://josev-api.switch-ev.com/external_messages/cs_status_and_limits

Then we need to define a new message, that is used to push the station limits/targets as mentioned here: https://www.notion.so/switchev/Load-Management-with-ISO-15118-2-and-wake-up-procedure-0cf256cf620e40119a184938ec43fc11#cbec90c7460f45ac995c741de0e22b42

I need to think on a good message name and structure to include all we need:
* schedule
* max/min power (active and reactive), current and voltage levels
* soc targets
* departure times

tropxy (changes requested):

iso15118/secc/controller/interface.py
```diff
@@ -566,6 +574,26 @@ async def get_evse_present_current(self) -> PVEVSEPresentCurrent:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    async def get_evse_present_voltage_v20(self) -> RationalNumber:
```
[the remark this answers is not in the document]
shalinnijel2: Have now modified the existing method to accept the protocol as well so it could return the right type. Have removed the above method.

iso15118/secc/controller/interface.py
```diff
@@ -590,6 +618,18 @@ async def start_cable_check(self):
         Relevant for:
         - DIN SPEC 70121
         - ISO 15118-2
+        - ISO 15118-20
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def get_cable_check_status(self) -> Processing:
```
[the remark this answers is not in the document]
shalinnijel2: Have modified the existing method in DINSPEC and -2 to use the new get_cable_check_status(), Have modified this method to return IsolationLevel though. The expectation is that method would return None if a status is not yet available.

iso15118/secc/controller/interface.py
shalinnijel2: Sure -  as discussed some of the methods have now been unified. This is a short summary of how the methods that have been updated:
```
get_ac_charge_params_v20()
get_ac_bpt_charge_params_v20()
==> get_ac_charge_params_v20(selected_service: ServiceV20)

get_dc_charge_params_v20()
get_dc_bpt_charge_params_v20()
==> get_dc_charge_params_v20(selected_service: ServiceV20)

get_scheduled_ac_charge_loop_params()
get_bpt_scheduled_ac_charge_loop_params()
get_dynamic_ac_charge_loop_params()
get_bpt_dynamic_ac_charge_loop_params()
==> get_ac_charge_loop_params_v20(control_mode, selected_service)

get_scheduled_dc_charge_loop_params()
get_bpt_scheduled_dc_charge_loop_params()
get_dynamic_dc_charge_loop_params()
get_bpt_dynamic_dc_charge_loop_params()
==> get_dc_charge_loop_params_v20(control_mode, selected_service)
```

tropxy: for historical reasons, I transcribe here the chat between I and Shalin:

> get_ac_charge_params_v20(selected_service: ServiceV20)
   get_dc_charge_params_v20(selected_service: ServiceV20)

those two we probably will get the values from the cs_status_and_limits
which we have to adapt to contain the max/min charge and discharge current/power values

I also realized something:
* In Dynamic DC Charge Loop (BPT or not) - The EVSE has full control over what the EV consumes at any point. This means that we dont need to change over time the EVSE limits provided to the EV. Those can be static since ChargeParameterDiscovery stage.
* In Scheduled DC Charge Loop (BPT or not) - The EVSE still has direct control over the power electronics. The difference here is that the EV is following a schedule, so it provides Targets to the EV based on the schedule. This is the equivalent DC use case of ISO 15118-2. In order to change the Current/Voltage requested by the EV we can either do a renegotiation or set the Max/Min limit of the station in the Res message
* In Dynamic AC Charge Loop (BPT or not) - The power electronics sits in the EV, so during this mode, the EV will tell the limits to the EVSE and the EVSE influences the EV with the Target Active and Reactive powers
* In Scheduled AC Charge Loop (BPT or not)  - The power electronics sits in the EV as well. This is the equivalent of the AC ISO15118-2 use case. The EV is just obligated to send the EVPresentActivePower and the EVSE can influence the schedule, either by renegotiation or by sending a Target


What is the interaction of Josev and the CS Hardware?
* In Dynamic DC Charge Loop (BPT or not): When Josev receives a new profile, it can directly send the new setpoint to the CS/power electronics of the DC station and no need to send anything to the ISO 15118
* In Scheduled DC Charge Loop (BPT or not): When Josev receives a new profile, it can directly send the new setpoint to the CS/power electronics of the DC station, but it may also need to change the Max/Min limits reported in ISO 15118 (if the setpoint deviates from the initial limits) or it can do a schedule renegotiation
* In Dynamic AC Charge Loop (BPT or not): When Josev receives a new profile, it  directly sends the new setpoint to the ISO 15118
* In Scheduled AC Charge Loop (BPT or not): When Josev receives a new profile, it  directly sends the new setpoint to the ISO 15118 or does a schedule renegotiation
When in basic charging [AC only]: when Josev has a new setpoint, it may use the cp_pwm message to change the duty cycle and influence the charging

iso15118/secc/states/iso15118_20_states.py
```diff
@@ -1450,7 +1476,66 @@ async def process_message(
...
+            # Any relays in the DC output circuit of the DC station shall
+            # be closed during the insulation test
+            if not await self.comm_session.evse_controller.is_contactor_closed():
+                self.stop_state_machine(
+                    "Contactor didnt close for Cable Check",
+                    message,
+                    ResponseCode.FAILED,
+                )
+                return
+            await self.comm_session.evse_controller.start_cable_check()
```
tropxy, line 1500: as I mentioned to you before, we need to invert this. First we start cable check, then we check if the contactors are closed

tropxy: I will approve the PR, so I dont block your work, but change that if you can ;)

tropxy (approved):

iso15118/secc/states/iso15118_20_states.py
shalinnijel2: Yes have created a ticket [here](https://dev.azure.com/switch-ev/Josev/_backlogs/backlog/Josev%20Team/Epics/?workitem=3669).

PR #180, opened 2022-12-16
Feature/optional ev data

martinbachmanndesignwerk:
Fixing unittest warnings
optional ev data from current demand added to EVDataContext


martinbachmanndesignwerk:
All changes here are contained in https://github.com/SwitchEV/iso15118/pull/232

PR #181, opened 2022-12-16
Feature/clean shutdown of session handler

martinbachmanndesignwerk:
Improvements to allow a comm session being started and stopped as part of another larger async application

### DIFF
--- a/iso15118/shared/utils.py
+++ b/iso15118/shared/utils.py
@@ -62,4 +62,4 @@ async def start_and_wait_for_coroutines(
     await_coroutines: List[Coroutine], return_when=asyncio.FIRST_EXCEPTION
 ):
     tasks = start_tasks(await_coroutines)
-    await wait_for_tasks(tasks)
+    await wait_for_tasks(tasks, return_when)


iso15118/shared/utils.py
```diff
@@ -49,3 +56,10 @@ async def wait_for_tasks(
             task.result()
         except Exception as e:
             logger.exception(e)
+
+
+async def start_and_wait_for_coroutines(
+    await_coroutines: List[Coroutine], return_when=asyncio.FIRST_EXCEPTION
+):
+    tasks = start_tasks(await_coroutines)
+    await wait_for_tasks(tasks)
```
tropxy, line 65: return when is not passed to the wait_for_tasks method

martinbachmanndesignwerk: good point. I've added a commit to fix that...

adoebber:
A clean shutdown for EVCC is still missing, isn't it? The `while True` of the `get_from_rcv_queue` function must be interrupted (missing `break` in the `elif isinstance(notification, StopNotification)` branch?). In addition `_transport` of the UDP client must be closed. Otherwise there is a runtime error.

PR #191, opened 2023-01-20
the name of is_external_authorization method has changed to is_eim_au…

ikaratass:
* The name of the is_external_authorization method is updated to is_eim_authorized to make it more clear.


mdwcrft:
How come the `poetry.lock` file is updated? This should not change unless dependency versions are changing

tropxy (approved):

shalinnijel2 (approved):

PR #204, opened 2023-03-02
added a way to report the charging power limits during the charging loop

tropxy:


iso15118/evcc/controller/simulator.py
```diff
@@ -177,18 +178,13 @@ async def select_energy_service_v20(
...
-        if matched_energy_services:
-            top_of_list: MatchedService = services[0]
-            selected_service = SelectedEnergyService(
-                service=top_of_list.service,
-                is_free=top_of_list.is_free,
-                parameter_set=top_of_list.parameter_sets[0],
-            )
-            return selected_service
-        return None
+        top_of_list: MatchedService = services[0]
```
shalinnijel2, line 181: Not sure if I'm missing something here. If there are multiple services offered by the secc and if the top service isn't energy service, wouldn't it fail after ServiceSelectionReq is received on the SECC side here: 
https://github.com/SwitchEV/iso15118/blob/159c6fc6e015c71d2744c044b877d97c3deba913/iso15118/secc/states/iso15118_20_states.py#L754

iso15118/secc/controller/simulator.py
```diff
@@ -578,6 +578,62 @@ async def get_evse_status(self) -> EVSEStatus:
     async def set_present_protocol_state(self, state_name: str):
         pass
 
+    async def send_charging_power_limits(
+        self,
+        protocol: Protocol,
+        control_mode: ControlMode,
+        selected__energy_service: ServiceV20,
```
shalinnijel2, line 585: Accidental double underscore for selected_energy_service?

iso15118/secc/controller/simulator.py
```diff
@@ -578,6 +578,62 @@ async def get_evse_status(self) -> EVSEStatus:
...
+        """
+        if protocol == Protocol.ISO_15118_20_AC:
+            if selected__energy_service in [ServiceV20.AC, ServiceV20.AC_BPT]:
+                charge_parameters = await self.get_ac_charge_params_v20(
+                    selected__energy_service
+                )
+            else:
+                charge_parameters = await self.get_dc_charge_params_v20(
+                    selected__energy_service
+                )
```
shalinnijel2, line 606: Do we expect this to be hit? Maybe once we add support for DC...if so the if condition above for protocol check could be modified to include ISO_15118_20_DC as well...

iso15118/secc/states/iso15118_20_states.py
```diff
@@ -1327,6 +1327,26 @@ async def process_message(
...
                     ControlMode.DYNAMIC, ServiceV20.AC_BPT
                 )  # noqa
+
+                # For now we just do this for the Dynamic Mode
+                ev_bpt_charge_parameters = ac_charge_loop_req.bpt_dynamic_params.dict()
+                # extract only the power limits and convert them to decimal
+                # representation
+                ev_power_limits = {}
+                for k, v in ev_bpt_charge_parameters.items():
+                    if v and v.get("exponent") is not None:
```
shalinnijel2, line 1337: Something minor - As the RationalNumber field here is received from the EV side, and as the field is not optional  - so I think the "if v and v.get("exponent")" check is being double safe but not required?

shalinnijel2 (changes requested):

iso15118/secc/controller/simulator.py
shalinnijel2, line 606: Do we expect the dc to be hit? Maybe once we add support for DC...if so the if condition above for protocol check could be modified to include ISO_15118_20_DC as well...

iso15118/evcc/controller/simulator.py
tropxy: but the condition that was there before wouldnt do anything useful. before we call this `select_energy_service_v20` method, we already have a guarantee that we have a matched service, so creating a list with just the matched energy services isnt anything new.

nevertheless, you are right, the first service of the SECC may not be an energy one and this fails, but anyway that is something to fix in another PR, I believe

iso15118/secc/states/iso15118_20_states.py
tropxy: you are right, the exponent check is not needed, but the value check it is, cause there are None values:

```
dict_items([('departure_time', None), ('ev_target_energy_request', {'exponent': 3, 'value': 40}), ('ev_max_energy_request', {'exponent': 3, 'value': 60}), ('ev_min_energy_request', {'exponent': 3, 'value': -20}), ('ev_max_charge_power', {'exponent': 3, 'value': 300}), ('ev_max_charge_power_l2', None), ('ev_max_charge_power_l3', None), ('ev_min_charge_power', {'exponent': 0, 'value': -100}), ('ev_min_charge_power_l2', None), ('ev_min_charge_power_l3', None), ('ev_present_active_power', {'exponent': 3, 'value': 200}), ('ev_present_active_power_l2', None), ('ev_present_active_power_l3', None), ('ev_present_reactive_power', {'exponent': 3, 'value': 20}), ('ev_present_reactive_power_l2', None), ('ev_present_reactive_power_l3', None), ('ev_max_discharge_power', {'exponent': 3, 'value': 11}), ('ev_max_discharge_power_l2', None), ('ev_max_discharge_power_l3', None), ('ev_min_discharge_power', {'exponent': -3, 'value': 1}), ('ev_min_discharge_power_l2', None), ('ev_min_discharge_power_l3', None), ('ev_max_v2x_energy_request', None), ('ev_min_v2x_energy_request', None)])
```

shalinnijel2 (approved):

PR #228, opened 2023-04-24
Fix: iso20 SessionStopReq incorrect payload type 0x8001

aadritG:
iso20 SessionStopReq is being sent with incorrect payload type 0x8001.


shalinnijel2:
Thank you @aadritG for your contribution... :)

PR #237, opened 2023-05-10
Fix tls_no_tls_switching_issue.

shalinnijel2:
- Fixed a TLS issue where once TLS server is created, a TCP server wouldn't be created when requested in the following session.

### DIFF
--- a/iso15118/secc/comm_session_handler.py
+++ b/iso15118/secc/comm_session_handler.py
@@ -309,7 +309,7 @@ class CommunicationSessionHandler:
 
     async def start_tcp_server(self, with_tls: bool):
         if self.tcp_server_handler and self.tcp_server.is_tls_enabled == with_tls:
-            return True
+            return
 
         if self.tcp_server_handler:
             logger.info("Reset current tcp handler.")
--- a/iso15118/secc/controller/simulator.py
+++ b/iso15118/secc/controller/simulator.py
@@ -504,6 +504,7 @@ class SimEVSEController(EVSEControllerInterface):
             sales_tariff_entry=sales_tariff_entries,
             num_e_price_levels=2,
         )
+
         # Putting the list of SAScheduleTuple entries together
         sa_schedule_tuple = SAScheduleTuple(
             sa_schedule_tuple_id=1,


tropxy:
Hey @shalinnijel2 check what I did in my PR for the TLS:
https://github.com/SwitchEV/iso15118/pull/233/files#diff-6c804a5c2a8823c6eb1ff47bd33cc864b7b1cb5bd40869bfa1a5bd5e16e4b79cL140

That should be equivalent to what you did, as you can check in the code here: https://github.com/python/cpython/blob/3.11/Lib/ssl.py#L745

we could just use the default context, but the fact is that we modify it later on for the EVCC side, also we shall not include the cafile attribute unless we want the mutual authentication. I also changed the env name, because we are not enabling TLS 1.3 with it...we are just enforcing the client to authenticate as well and that can be done in 1.2. The 1.3 Vs 1.2 will depend on the cipher suits that the client and server support... 

In theory we can even use the -2 certificates for TLS 1.3 connection. let me know your thoughts about this

tropxy:
> The env name change makes sense. But there is one thing that is not clear to me yet. As the pki for -20 uses a different curve, how do we handle both -2 and -20 at the same time? Would we need another env for that if it can't be handled automatically?

so, that is a tricky one... we dont know if the connection is a -20 or -2 one until we are in supportedAppProtocol, which is then too late for the TLS handshake; I dont think we can actively control this if we want to support -2 and -20 at the same time....so I think the best here is to support all the groups by default.

i did some tests, and the Server will pick the client first supported named group in the Client Hello, that the Server also supports. Maybe in SupportedAppProtocol, we can later detect which was the agreed curve, and kill the connection if the curve is not a -20 one, but I think this may be complicated. I also tried to use the `set_ecdh_curve` and that sets one curve that the server supports and if we call it again it overrides the previous call, which means we cant set multiple curves with that function and there is not another one that we can use in python for that purpose.

I think for the best is really to not set any particular curve in the server and let client and server handle it between them. Once we have the TLS from Louis, maybe we can do some more sophisticated settings.

iso15118/secc/comm_session_handler.py
```diff
@@ -308,11 +308,16 @@ async def end_current_session(self, peer_ip_address: str):
         self.udp_server.resume_udp_server()
 
     async def start_tcp_server(self, with_tls: bool):
-        if self.tcp_server_handler is not None:
-            """A TCP server is already available, ready to respond.
-            (Perhaps created when the last SDP request was received.)
-            """
-            return
+        if self.tcp_server_handler and self.tcp_server.is_tls_enabled == with_tls:
+            return True
```
ikaratass, line 312: do we need to return it as True, it seems just 'return'

shalinnijel2: yes, fixed.

ikaratass (approved):

PR #241, opened 2023-06-07
AB#4965 : Fixed dummy schedule generation

shalinnijel2:
Fixed dummy schedule generation to cope with departure_time greater than 172800 days.


ikaratass (approved):
LGTM

PR #249, opened 2023-06-16
AB#5093 Skip SalesTariff if free service.

shalinnijel2:
If free service is set to true, then no SalesTariff is set in charging schedules.


ikaratass (approved):
LGTM

PR #265, opened 2023-06-27
AB#5144:support to return certificate validation status.

shalinnijel2:
Return certificate validation result if available.


ikaratass (approved):
LGTM

PR #266, opened 2023-06-30
Python gc

ikaratass:
* Garbage collector is disabled during processing message


iso15118/shared/comm_session.py
```diff
@@ -545,3 +557,6 @@ async def rcv_loop(self, timeout: float):
                 await self.stop(stop_reason)
                 self.session_handler_queue.put_nowait(self.stop_reason)
                 return
+            finally:
+                if gc_was_enabled:
+                    gc.disable()
```
shalinnijel2, line 562: gc.enable() ?

shalinnijel2 (changes requested):

shalinnijel2 (approved):

PR #274, opened 2023-07-10
Make running of udp server optional

shalinnijel2:
Running of UDP server is optional. Default behaviour stays the same as before.

### DIFF
--- a/iso15118/secc/__init__.py
+++ b/iso15118/secc/__init__.py
@@ -26,10 +26,10 @@ class SECCHandler(CommunicationSessionHandler):
             evse_controller,
         )
 
-    async def start(self, iface: str):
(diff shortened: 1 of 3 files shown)


santiagosalamandri (approved):
LGTM

GUANMINLIAO:
Hello,
     But could I running udp_client optionally? I want to assign ip address and port manually, many thanks!

shalinnijel2:
Yes, UDP server could be disabled by setting the optional argument, start_udp_server to `False`. 

Once this is done, the following method could be called to process the received SDP request which will then setup the tcp/tls server and return details as an SDP payload. 
https://github.com/SwitchEV/iso15118/blob/94f41d65911314185d8aa3453321228b7bd1620f/iso15118/secc/comm_session_handler.py#L364

PR #279, opened 2023-07-21
Updates to store EV and EVSE limits shared in CPD and CL (-20 AC and DC).

shalinnijel2:
In this PR:
1. Save the ev/evse limits shared during AC/DC ChargeParameterDiscoveryReq/Res (tests added)
2. Update/read these limits during charging loop. (tests added)


iso15118/secc/controller/simulator.py
```diff
@@ -722,30 +722,66 @@ async def get_ac_charge_params_v20(
...
             return BPTACChargeParameterDiscoveryResParams(
                 **(ac_charge_parameter_discovery_res_params.dict()),
-                evse_max_discharge_power=RationalNumber(exponent=0, value=3000),
-                evse_max_discharge_power_l2=RationalNumber(exponent=0, value=3000),
-                evse_max_discharge_power_l3=RationalNumber(exponent=0, value=3000),
-                evse_min_discharge_power=RationalNumber(exponent=0, value=300),
-                evse_min_discharge_power_l2=RationalNumber(exponent=0, value=300),
-                evse_min_discharge_power_l3=RationalNumber(exponent=0, value=300),
+                evse_max_discharge_power=RationalNumber.get_rational_repr(
+                    self.evse_data_context.evse_max_discharge_power_l3
```
ikaratass, line 768: self.evse_data_context.evse_max_discharge_power_l3 -> should be  self.evse_data_context.evse_max_discharge_power

shalinnijel2: Yes, have fixed in now...

ikaratass (dismissed):
LGTM

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,152 @@
...
+
+    ev_max_v2x_energy_request: Optional[float] = None
+    ev_min_v2x_energy_request: Optional[float] = None
+
+    # Seen in Scheduled DC CL
+    ev_target_current: Optional[float] = None
+    ev_target_voltage: Optional[float] = None
+    ev_max_charge_power: Optional[float] = None
+    ev_min_charge_power: Optional[float] = None
+    ev_max_charge_current: Optional[float] = None
```
tropxy, line 95: repeated here: https://github.com/SwitchEV/iso15118/pull/279/files#diff-4100705728a97ab69df949905fd16a36dbed66eb00d125c361894c1e8356b9d8R47

tropxy, line 95: repeated here: https://github.com/SwitchEV/iso15118/pull/279/files#diff-4100705728a97ab69df949905fd16a36dbed66eb00d125c361894c1e8356b9d8R67

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,152 @@
...
+    # Specific to ISO 15118-20 DC
+    ev_max_charge_current: Optional[float] = None
+    ev_min_charge_current: Optional[float] = None
+    ev_max_voltage: Optional[float] = None
+    ev_min_voltage: Optional[float] = None
+    target_soc: Optional[int] = None
+
+    # Specific to ISO 15118-20 DC BPT
+    ev_max_discharge_current: Optional[float] = None
+    ev_min_discharge_current: Optional[float] = None
```
tropxy, line 75: I dont think EV min discharge current exists

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,152 @@
...
+    ev_present_reactive_power_l2: Optional[float] = None
+    ev_present_reactive_power_l3: Optional[float] = None
+
+    ev_max_v2x_energy_request: Optional[float] = None
+    ev_min_v2x_energy_request: Optional[float] = None
+
+    # Seen in Scheduled DC CL
+    ev_target_current: Optional[float] = None
+    ev_target_voltage: Optional[float] = None
+    ev_max_charge_power: Optional[float] = None
```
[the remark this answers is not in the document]
shalinnijel2: Yes, removed now.

iso15118/secc/controller/ev_data.py
shalinnijel2: They are present in DC CPD: https://github.com/SwitchEV/iso15118/blob/5b048f90a9fcb0aa4f0310ffa9be943c729ee37e/iso15118/shared/messages/iso15118_20/dc.py#L73

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,152 @@
...
+
+    ev_target_energy_request: Optional[float] = None
+    ev_max_energy_request: Optional[float] = None
+    ev_min_energy_request: Optional[float] = None
+    ev_present_active_power: Optional[float] = None
+    ev_present_active_power_l2: Optional[float] = None
+    ev_present_active_power_l3: Optional[float] = None
+    ev_present_reactive_power: Optional[float] = None
+    ev_present_reactive_power_l2: Optional[float] = None
+    ev_present_reactive_power_l3: Optional[float] = None
```
tropxy, line 85: this exists but only in the context dynamic AC charging loop

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,152 @@
...
+
+    ev_max_v2x_energy_request: Optional[float] = None
+    ev_min_v2x_energy_request: Optional[float] = None
+
+    # Seen in Scheduled DC CL
+    ev_target_current: Optional[float] = None
+    ev_target_voltage: Optional[float] = None
+    ev_max_charge_power: Optional[float] = None
+    ev_min_charge_power: Optional[float] = None
```
tropxy: repeated here: 
https://github.com/SwitchEV/iso15118/pull/279/files#diff-4100705728a97ab69df949905fd16a36dbed66eb00d125c361894c1e8356b9d8R48

iso15118/secc/controller/ev_data.py
tropxy: yes, sorry, I meant to add in a review and not as single comment and I deleted my comment when I realized I was wrong

iso15118/secc/controller/ev_data.py
shalinnijel2: Yes, removed now.

iso15118/secc/controller/ev_data.py
shalinnijel2: Yes, have now added a comment to clarify this.

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,214 @@
...
+        for k, v in params.items():
+            if type(v) is dict:
+                ev_params_dict.update({k: v["value"] * 10 ** v["exponent"]})
+            elif type(v) in [int, float]:
+                ev_params_dict.update({k: v})
+
+        self.__dict__.update(ev_params_dict)
+
+    def as_dict(self):
+        return self.__dict__
```
tropxy, line 86: didnt we have a base class from this one could inherit these methods from?

shalinnijel2:
Hmm...I don't think there is one. You did define a method to do this in
josev_pro land I think...

iso15118/secc/states/iso15118_20_states.py
```diff
@@ -1769,7 +1795,39 @@ async def process_message(
...
+            if control_mode == ControlMode.SCHEDULED:
+                self.comm_session.evse_controller.ev_data_context.ev_session_context.update(  # noqa
+                    dc_charge_loop_req.scheduled_params
+                )
+            elif control_mode == ControlMode.DYNAMIC:
+                self.comm_session.evse_controller.ev_data_context.ev_session_context.update(  # noqa
+                    dc_charge_loop_req.dynamic_params
+                )
+        elif selected_energy_service.service == ServiceV20.DC_BPT:
+            if control_mode == ControlMode.SCHEDULED:
```
tropxy, line 1814: it may be a bit of a hacky/obfuscated way to do this, but what if we do:
```python
# this gets the scheduled or dynamic function attribute
bpt_attributes = getattr(dc_charge_loop_req, 'bpt_' + control_mode.name.lower())
self.comm_session.evse_controller.ev_data_context.ev_session_context.update( 
                    bpt_attributes
```

It saves some lines of code, but maybe turns the code less readable and prone to error if in the future we change the attributes names. What do you think?

tropxy (changes requested):

iso15118/secc/controller/ev_data.py
shalinnijel2: Hmm...I don't think there is one. You did define a method to do this in
josev_pro land I think...

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,214 @@
...
+    ScheduledDCChargeLoopReqParams,
+)
+
+
+@dataclass
+class EVRatedLimits:
+    # Common to both ISO15118-20 AC and DC
+    ev_max_charge_power: Optional[float] = 0.0
+    ev_min_charge_power: Optional[float] = 0.0
+
```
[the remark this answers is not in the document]
shalinnijel2: They are present here - https://github.com/SwitchEV/iso15118/blob/a86ccefbb4ffe8d6849498e8bee950813ef9e83c/iso15118/secc/controller/ev_data.py#L52-L53

iso15118/secc/controller/ev_data.py
shalinnijel2: Oh yes there is... :)...
Here: https://github.com/SwitchEV/iso15118/blob/a86ccefbb4ffe8d6849498e8bee950813ef9e83c/iso15118/secc/controller/evse_data.py#L6-L21

shalinnijel2: But I accidentally added the inheritance only for the ones in evse_data. Will add them for ev_data as well...

shalinnijel2: Have moved this to a common interface now... https://github.com/SwitchEV/iso15118/blob/dc_bpt_change/iso15118/secc/controller/common.py

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,214 @@
...
+    ev_max_charge_power: Optional[float] = None
+    ev_min_charge_power: Optional[float] = None
+
+    # Specific to ISO 15118-20 AC
+    ev_max_charge_power_l2: Optional[float] = None
+    ev_max_charge_power_l3: Optional[float] = None
+    ev_min_charge_power_l2: Optional[float] = None
+    ev_min_charge_power_l3: Optional[float] = None
+
+    # Specific to ISO 151180-20 Dynamic AC CL
```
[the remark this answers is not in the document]
shalinnijel2: Yes, have added a comment to clarify...

iso15118/secc/controller/evse_data.py
```diff
@@ -0,0 +1,143 @@
...
+    #  Optional in 15118-2 CPD
+    evse_current_regulation_tolerance: Optional[float] = None
+    evse_peak_current_ripple: Optional[float] = None
+    evse_energy_to_be_delivered: Optional[float] = None
+
+
+@dataclass
+class EVSERatedLimits:
+    ac_limits: Optional[ACLimits] = None
+    ac_bpt_limits: Optional[ACBPTLimits] = None
```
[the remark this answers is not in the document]
shalinnijel2: My thinking was as cs limits splits these, this would make it easier for implementers. This would make it evident that the bpt bit could be skipped if not interesting.

iso15118/secc/controller/evse_data.py
```diff
@@ -0,0 +1,143 @@
...
+    dc_limits: Optional[DCLimits] = None
+    dc_bpt_limits: Optional[DCBPTLimits] = None
+
+
+@dataclass
+class EVSESessionContext:
+    # Optional in -20 Dynamic CL Res
+    departure_time: Optional[int] = None
+    min_soc: Optional[int] = None
+    target_soc: Optional[int] = None
```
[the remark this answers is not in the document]
shalinnijel2: Yes, sounds good. Have made the changes.

iso15118/secc/states/iso15118_20_states.py
shalinnijel2: Yea - right now even though there are a few more lines, what is expected is evident. The above change would break if we change the field names like you said...

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,214 @@
...
+        self.__dict__.update(ev_params_dict)
+
+    def as_dict(self):
+        return self.__dict__
+
+
+@dataclass
+class EVSessionContext:
+    dc_current_request: Optional[int] = None
+    dc_voltage_request: Optional[int] = None
```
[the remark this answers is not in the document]
shalinnijel2: Yea - the other fields are currently used by -2 and din. I'll refactor once the -20 bit is done. Would that be ok?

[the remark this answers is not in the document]
tropxy: ok, no worries

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,214 @@
...
+    ev_max_discharge_power: Optional[float] = None
+    ev_min_discharge_power: Optional[float] = None
+
+    ev_max_v2x_energy_request: Optional[float] = None
+    ev_min_v2x_energy_request: Optional[float] = None
+
+    # Seen in Scheduled DC CL
+    ev_target_current: float = 0.0
+    ev_target_voltage: float = 0.0
+    ev_min_charge_current: Optional[float] = None
```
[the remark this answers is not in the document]
tropxy: I meant, the value exist but in the CPD, not in the CL

iso15118/secc/controller/evse_data.py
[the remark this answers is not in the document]
shalinnijel2: Yes, if picking the bpt version, the non-bpt equivalent will also be required.
We can't make ac_limits non optional here as dc_limits also lives in the same RatedLimits dataclass.
One option is to let the bpt limits live as optional ones in the ac_limits/dc_limits dataclass. But splitting them makes it slightly easier to read I think. But as these rated limits are built from cs-config we can clarify (will have to be in the documentation though) that this is needed. Hmm, do we need add a __post_init__() to validate?

iso15118/secc/controller/ev_data.py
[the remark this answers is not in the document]
shalinnijel2: Yes, I have split the params into CPD and CL related ones like we did for EVSE to help make this clearer.

iso15118/secc/controller/evse_data.py
```diff
@@ -0,0 +1,126 @@
...
+
+@dataclass
+class EVSEACBPTLimits(Limits):
+    evse_max_discharge_power: Optional[float] = None  # Required
+    evse_max_discharge_power_l2: Optional[float] = None  # Optional
+    evse_max_discharge_power_l3: Optional[float] = None  # Optional
+
+    evse_min_discharge_power: Optional[float] = None  # Required
+    evse_min_discharge_power_l2: Optional[float] = None  # Optional
+    evse_min_discharge_power_l3: Optional[float] = None  # Optional
```
tropxy, line 53: do we actually need to segregate these two classes? 
why dont we just have the class `EVSEACCPDLimits` and add a comment `# BPT limits` and below the `evse_..._discharge..` fields?

But if we need the two classes, then shall we not subclass the `EVSEACBPTCPDLimits` from the `EVSEACCPDLimits` class?

iso15118/secc/controller/evse_data.py
```diff
@@ -0,0 +1,126 @@
...
+    evse_target_reactive_power: Optional[float] = None
+    evse_target_reactive_power_l2: Optional[float] = None
+    evse_target_reactive_power_l3: Optional[float] = None
+    evse_present_active_power: Optional[float] = None  # Optional in AC CPD
+    evse_present_active_power_l2: Optional[float] = None  # Optional in AC CPD
+    evse_present_active_power_l3: Optional[float] = None  # Optional in AC CPD
+
+
+@dataclass
+class EVSEACLimits(Limits):
```
tropxy, line 22: can we change the name to `EVSEACCPDLimits` and add a description to the class expanding the name

iso15118/secc/controller/evse_data.py
```diff
@@ -0,0 +1,126 @@
...
+    evse_min_voltage: Optional[float] = None  # Required in 15118-20 Dynamic BPT CL
+
+
+@dataclass
+class EVSEDCBPTLimits(Limits):
+    # Required in 15118-20 DC BPT CPD
+    evse_max_discharge_power: Optional[float] = None
+    evse_min_discharge_power: Optional[float] = None
+    evse_max_discharge_current: Optional[float] = None
+    evse_min_discharge_current: Optional[float] = None
```
tropxy, line 77: same comment as before, cant we just have 1 class?

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,187 @@
...
+    ev_min_discharge_power_l2: Optional[float] = None
+    ev_min_discharge_power_l3: Optional[float] = None
+
+
+@dataclass
+class EVDCLimits(Limits):
+    ev_max_charge_power: Optional[float] = 0.0
+    ev_min_charge_power: Optional[float] = 0.0
+    ev_max_charge_current: Optional[float] = None
+    ev_min_charge_current: Optional[float] = None
```
tropxy, line 100: minor detail: as the CPD comes first than the CL, can we move this class above the the other 2?

also, can we change the names to `EVACCPDLimits` and `EVDCCPDLimits` and add a description where at least the names are expanded?

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(
+        self,
+        ac_limits: Optional[EVACLimits] = None,
+        dc_limits: Optional[EVDCLimits] = None,
+    ):
+        self.ac_limits = ac_limits or EVACLimits()
+        self.dc_limits = dc_limits or EVDCLimits()
+
+    ac_limits: Optional[EVACLimits] = None
+    dc_limits: Optional[EVDCLimits] = None
```
tropxy, line 122: what is the purpose of these ones?

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,187 @@
...
+    dc_voltage_request: Optional[int] = None
+    ac_current: Optional[dict] = None  # {"l1": 10, "l2": 10, "l3": 10}
+    ac_voltage: Optional[dict] = None  # {"l1": 230, "l2": 230, "l3": 230}
+    soc: Optional[int] = None  # 0-100
+
+    remaining_time_to_full_soc_s: Optional[float] = None
+    remaining_time_to_bulk_soc_s: Optional[float] = None
+
+    ac_limits: Optional[EVACCLLimits] = None
+    dc_limits: Optional[EVDCCLLimits] = None
```
tropxy, line 145: what is the purpose of these ones?

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,187 @@
...
+        ev_rated_limits: Optional[EVRatedLimits] = None,
+        ev_session_context: Optional[EVSessionContext] = None,
+    ):
+        self.evcc_id = evcc_id or None
+        self.ev_rated_limits = ev_rated_limits or EVRatedLimits()
+        self.ev_session_context = ev_session_context or EVSessionContext()
+
+    evcc_id: Optional[str] = None
+    ev_rated_limits: Optional[EVRatedLimits] = None
+    ev_session_context: Optional[EVSessionContext] = None
```
tropxy, line 162: and these ones?

iso15118/shared/messages/iso15118_20/common_types.py
```diff
@@ -174,6 +174,125 @@ class RationalNumber(BaseModel):
...
+        # This is to ensure that we keep the value to be converted, within
+        # the range of the classes RationalNumber and PhysicalValue 'value' data type
+        # int16
+        min_limit = max(min_limit, INT_16_MIN)
+        # For example, the class PVEVSEMaxPowerLimit has a _max_limit = 200000,
+        # so, if the value to be converted is 200000, we cant represent it
+        # with exponent = 0; instead, we need to take into account
+        # the INT_16_MAX = 32768 and represent the value as
+        # value = 20000, exponent = 1
+        max_limit = min(max_limit, INT_16_MAX)
```
tropxy, line 243: we will have trouble with my code in https://github.com/SwitchEV/iso15118/pull/319/files#diff-dffae5111fad8ea791bc869910a9e31c6cd21e49ab7c89bc7bc6b6ad36a8ab2bR51

I guess it is better if I set the limits to the max representations of it in my code

iso15118/shared/messages/iso15118_20/common_types.py
```diff
@@ -174,6 +174,125 @@ class RationalNumber(BaseModel):
...
+        if abs_value < limit:
+            while abs_value < limit and exponent > -3:
+                exponent -= 1
+                abs_value = abs_value * 10
+            if abs_value > limit:
+                # for example for a value of 2.34 and max limit = 400
+                # with exponent = -3 we would send an end value of 2340,
+                # which does not fit the limit, so we go back one exponent
+                # to be able to satisfy the type limitation:
+                # exponent = -2 and end value = 234
```
tropxy, line 275: I think I found a better algorithm:

```python
import math
def float_to_int_and_exponent(
    value: float,
    min_limit: float = INT_16_MIN,
    max_limit: float = INT_16_MAX
) -> Tuple[int, int]:

    # if the value is 0 return immediately 
    if value == 0:
        return 0, 0

    # Check the sign of the value to determine how to round the result
    round_func = math.floor if value > 0 else math.ceil

    # Iterate over possible exponents to find the best approximation
    for exponent in range(-3, 4):
        new_value = round_func(value * 10 ** (-exponent))
        # If new_value after rounding fits within the range of a 16-bit integer, return it
        if min_limit <= new_value <= max_limit:
            return new_value, exponent

    # If no exponent could make the value fit in a 16-bit integer range,
    # return the original value with exponent 0
    return round_func(value), 0
```

tropxy (changes requested):

iso15118/secc/controller/ev_data.py
shalinnijel2: Sure, done.

iso15118/secc/controller/evse_data.py
shalinnijel2: These will be direct mapping from cs_limits in pro land. So for now, I have kept the classes separate but have changed the BPT versions to inherit from the respective class.

iso15118/secc/controller/evse_data.py
shalinnijel2: Sure, done.

iso15118/secc/controller/evse_data.py
shalinnijel2: As mentioned above, have changed the BPT version to inherit from the "non-bpt" one.

iso15118/shared/messages/iso15118_20/common_types.py
shalinnijel2: Hmm..this is however specific to -20. The change you added would apply only to -2 and din, right?

tropxy: that is true...we dont use the convert_to_exponent number there do we?
not sure why we dont tho lol

iso15118/shared/messages/iso15118_20/common_types.py
shalinnijel2: Ok, replaced the method with this now. I moved the test for this over from josev_pro land and it passed.

iso15118/secc/controller/ev_data.py
shalinnijel2: Done - removed.

iso15118/secc/controller/ev_data.py
shalinnijel2: Done - removed.

iso15118/secc/controller/ev_data.py
shalinnijel2: Done - removed.

iso15118/secc/states/iso15118_20_states.py
```diff
@@ -1769,7 +1795,39 @@ async def process_message(
...
+                self.comm_session.evse_controller.ev_data_context.ev_session_context.dc_limits.update(  # noqa
+                    dc_charge_loop_req.dynamic_params.dict()
+                )
+        elif selected_energy_service.service == ServiceV20.DC_BPT:
+            if control_mode == ControlMode.SCHEDULED:
+                self.comm_session.evse_controller.ev_data_context.ev_session_context.dc_limits.update(  # noqa
+                    dc_charge_loop_req.bpt_scheduled_params.dict()
+                )
+            else:
+                self.comm_session.evse_controller.ev_data_context.ev_session_context.dc_limits.update(  # noqa
```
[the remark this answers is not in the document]
shalinnijel2: Yes, refactored.

iso15118/secc/controller/ev_data.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@dataclass
+class EVChargeParamsLimits:
+    ev_max_voltage: Optional[Union[PVEVMaxVoltageLimit, PVEVMaxVoltage]] = None
+    ev_max_current: Optional[Union[PVEVMaxCurrentLimit, PVEVMaxCurrent]] = None
+    ev_max_power: Optional[PVEVMaxPowerLimit] = None
+    e_amount: Optional[PVEAmount] = None
+    ev_energy_request: Optional[PVEVEnergyRequest] = None
```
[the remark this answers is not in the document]
tropxy: I understand that this class is closer to the -2 attributes, but I dont like we have duplicated data in different classes... like we repeat max voltage and power and e_amount which only appears in AC but is equivalent to ev_target_energy_request

We should have EV data that encapsulates all we need from the different states of the protocol but not be so tight to it that makes us duplicate data... If I would be a new user, I would get confused about it

tropxy (changes requested):

iso15118/secc/states/iso15118_20_states.py
```diff
@@ -1760,7 +1771,22 @@ async def process_message(
...
+        dc_charge_loop_req: DCChargeLoopReq = cast(DCChargeLoopReq, msg)
+
+        self.update_dc_charge_loop_params(
+            dc_charge_loop_req, selected_energy_service, control_mode
+        )
+        await self.comm_session.evse_controller.send_charging_power_limits(
+            self.comm_session.protocol,
+            control_mode,
+            selected_energy_service.service,
+        )
```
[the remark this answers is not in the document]
shalinnijel2: Yes, wrapped in try catch. The idea at the moment being any sort of error would be caught in the interface implementation, re-thrown as a timeout exception and the reason passed to the EV.

tropxy (approved):

PR #282, opened 2023-07-27
JTT-52: Stop charging if state C not detected in PowerDelivery state.

shalinnijel2:
[V2G20 - 847] The EVCC shall signal CP State C or D no later than 250 ms after sending the first PowerDeliveryReq with ChargeProgress equals "Start" within V2G communication session.


ikaratass (approved):
LGTM

PR #284, opened 2023-07-31
Simulate precharge ongoing before finished

shalinnijel2:
Simulate precharge for 5 cycles before returning finished.


ikaratass (approved):
LGTM

PR #285, opened 2023-08-01
Simulate welding detection ongoing before finished

shalinnijel2:
Updated ev simulator to simulate ongoing WeldingDetection before returning finished.


ikaratass (approved):
LGTM

PR #286, opened 2023-08-03
Return FAILED_CertificateExpired is certificate is not yet valid

shalinnijel2:
[V2G2-824] : The message 'PaymentDetailsRes' shall contain the ResponseCode 'FAILED_CertificateExpired' if the Contract Certificate contained in the PaymentDetailsReq message in attribute ContractSignatureCertChain is not yet valid.


ikaratass (approved):
LGTM

PR #289, opened 2023-08-15
JTT-81 Feat: Check response code

santiagosalamandri:
This PR introduces a status code verification before sending the response, as the compatibility might be affected by the state being transmitted.

**Note:** 
This limitation arises from the definition of the status response code and its alignment with Table 112 in iso15118-2, which constrains the response to match its corresponding state.


iso15118/secc/states/iso15118_2_states.py
```diff
@@ -970,8 +970,29 @@ async def process_message(
...
+                        ResponseCode.FAILED,
+                        ResponseCode.FAILED_SEQUENCE_ERROR,
+                        ResponseCode.FAILED_SIGNATURE_ERROR,
+                        ResponseCode.FAILED_UNKNOWN_SESSION,
+                        ResponseCode.FAILED_CERTIFICATE_EXPIRED,
+                        ResponseCode.FAILED_CERTIFICATE_REVOKED,
+                        ResponseCode.FAILED_NO_CERTIFICATE_AVAILABLE,
+                    ]
+                    else ResponseCode.FAILED
+                )
```
shalinnijel2, line 995: This could potentially change in an upcoming update to the spec based on the ticket here: https://iso15118.elaad.io/pt2/15118-2/user-group/-/issues/38
Also André has opened a ticket for a better response code when chain error occurs here: https://iso15118.elaad.io/pt2/15118-2/user-group/-/issues/41

PR #292, opened 2023-08-18
ScheduleExchange(StateEVCC_bug)

GUANMINLIAO:
Dear Authors:
       I change class ScheduleExchange(StateEVCC)   
       
       if self.comm_session.selected_energy_service in (
                ServiceV20.AC_BPT,
                ServiceV20.DC_BPT,
            ):
            
       to
       
       if self.comm_session.selected_energy_service.service in (
              ServiceV20.AC_BPT,
              ServiceV20.DC_BPT,
          ):
Otherwise, ChannelSelection.DISCHARGE or ChannelSelection.CHARGE cannot be determined.

Many thanks,
Mike

### DIFF
--- a/iso15118/evcc/states/iso15118_20_states.py
+++ b/iso15118/evcc/states/iso15118_20_states.py
@@ -841,7 +841,7 @@ class ScheduleExchange(StateEVCC):
 
             # Information from EV to show if charging or discharging is planned
             bpt_channel_selection = None
-            if self.comm_session.selected_energy_service in (
+            if self.comm_session.selected_energy_service.service in (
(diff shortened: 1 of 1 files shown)


shalinnijel2 (approved):
I've approved the PR as the fix is required to add the bpt_channel information in PowerDeliveryReq. However, I'm not still not very clear on how the information will be used. This is what I see in the spec:
```
[V2G20-2123] If the BPT service was selected in either AC or DC and the reverse power transfer system requires HLC-based control of switching electricity power channels, the parameter BPT_ChannelSelection of PowerDeliveryReq shall be applied.
```

PR #294, opened 2023-09-07
JTT-112 Feat: Enable mypy support

santiagosalamandri:
This PR includes:
- Resolution of all mypy, black, isort, and flake-related issues.
- Mypy enabled in code-quality checks.
- Integration of Mypy into the GitHub Actions workflow for automated checks.

**Validation**
Testing was carried out on the SECC/EVCC simulator, resulting in a clean execution with no issues, warnings, or exceptions. for the following protocols/modes:

| Protocol   | Status |
|:-----------|:-------|
| DIN        |  PASS  | 
|-2 ac eim   |  PASS  |
|-2 dc eim   |  PASS  |
|-2 ac pnc   |  PASS  |
|-2 dc pnc   |  PASS  |
|-20 ac      |  PASS  |
|-20 dc      |  PASS  |
|-20 ac-bpt  |  PASS  |
|-20 dc-bpt  |  PASS  |


iso15118/shared/utils.py
```diff
@@ -34,8 +34,9 @@ def load_requested_protocols(read_protocols: Optional[List[str]]) -> List[Protoc
             f"No supported protocols configured. Supported protocols are "
             f"{supported_protocols} and could be configured in evcc_config.json"
         )
-    supported_protocols = [Protocol[x] for x in valid_protocols]
-    return supported_protocols
+    return [
+        Protocol[name] for name in supported_protocols if name in Protocol.__members__
+    ]
```
shalinnijel2, line 39:
```suggestion
    return [
        Protocol[name] for name in valid_protocols if name in Protocol.__members__
    ]
```
This should be valid_protocols . As else the full list would be returned?

shalinnijel2 (changes requested):

iso15118/shared/states.py
```diff
@@ -338,7 +342,7 @@ def create_next_message(
                 exi_payload = EXI().to_exi(to_be_exi_encoded, namespace)
 
                 if hasattr(self.comm_session, "evse_id"):
-                    logger.trace(
+                    logger.debug(
```
shalinnijel2, line 345: This should be set back to .trace()

iso15118/shared/comm_session.py
```diff
@@ -203,25 +203,18 @@ async def process_message(self, message: bytes):
             )
 
             if hasattr(self.comm_session, "evse_id"):
-                logger.trace(
+                logger.debug(
```
shalinnijel2, line 206: This should be set back to .trace()

shalinnijel2 (changes requested):

iso15118/shared/security.py
```diff
@@ -532,9 +533,15 @@ def verify_certs(
...
-        ).value.path_length
+        try:
+            basic_contrains = cert.extensions.get_extension_for_oid(
+                ExtensionOID.BASIC_CONSTRAINTS
+            ).value
+            path_len = 0
+            if isinstance(basic_contrains, extensions.BasicConstraints):
+                path_len = basic_contrains.path_length
+        except ExtensionNotFound:
+            path_len = None
```
shalinnijel2, line 544: I think if we set path_len to None, we risk raising another exception while trying to raise CertChainLengthError below. as num_sub_cas is an int. Is it better to modify the definition for that to include another message field and making num_sub_cas optional?

shalinnijel2 (changes requested):

iso15118/secc/controller/simulator.py
```diff
@@ -1069,24 +1107,33 @@ async def get_15118_ev_certificate(
...
-        to_be_exi_encoded = V2GMessageV2(header=header, body=body)
-        exi_encoded_cert_installation_res = EXI().to_exi(
-            to_be_exi_encoded, Namespace.ISO_V2_MSG_DEF
-        )
+        if isinstance(cert_install_req, SupportedAppProtocolReq) or isinstance(
+            cert_install_req, SupportedAppProtocolRes
+        ):
+            logger.info(f"Ignoring EXI decoding of a {type(cert_install_req)} message.")
+            return ""
+        else:
```
shalinnijel2, line 1115: This if block is not required. Everything in the else condition is required though. :)

shalinnijel2 (changes requested):

iso15118/secc/controller/simulator.py
santiagosalamandri: @shalinnijel2 : `cert_install_req` is derived from the` EXI().from_exi()` function, which has the potential to yield various data types, some of which may lack the associated header field. This absence could result in a program error when accessed. Consequently, removing the conditional filter for variable type could lead to mypy raising an error, specifically stating the absence of a "header" field.

iso15118/shared/security.py
santiagosalamandri: Good catch! I think I tried to set it as "0" instead of None, so the next piece of code would raise the "CertAttributeError" exception. Either way, I'll raise that exception directly.

iso15118/shared/states.py
santiagosalamandri: Done

iso15118/shared/comm_session.py
santiagosalamandri: Done

iso15118/shared/utils.py
santiagosalamandri: that's correct. Fixed

iso15118/secc/controller/simulator.py
shalinnijel2: @santiagosalamandri - ok, understood. Would something like this be slightly better?
So we still do type checking to satisfy mypy as follows:
```
        if isinstance(cert_install_req, CertificateInstallationReq):
            header = MessageHeaderV2(
                session_id=cert_install_req.header.session_id,
                signature=signature,
            )
            body = Body.parse_obj(
                {"CertificateInstallationRes": cert_install_res.dict()}
            )
            to_be_exi_encoded = V2GMessageV2(header=header, body=body)
            exi_encoded_cert_installation_res = EXI().to_exi(
                to_be_exi_encoded, Namespace.ISO_V2_MSG_DEF
            )

            # base64.b64encode in Python is a binary transform
            # so the return value is byte[]
            # But the CPO expects exi_encoded_cert_installation_res
            # as a string, hence the added .decode("utf-8")
            base64_encode_cert_install_res = base64.b64encode(
                exi_encoded_cert_installation_res
            ).decode("utf-8")

            return base64_encode_cert_install_res
        else:
            logger.info(f"Ignoring EXI decoding of a {type(cert_install_req)} message.")
            return ""
```

shalinnijel2 (approved):

PR #297, opened 2023-09-12
Add bpt channel info only if evpowerprofile is available

shalinnijel2:
Check if ev processing if finished before accessing ev_power_profile for populating bpt_channel information.


santiagosalamandri (approved):
LGTM

PR #302, opened 2023-09-20
Simplify instructions to run locally

OrangeTux:
Starting the python scripts using `poetry run python <script>` instead of `python <script>`. The call to `poetry run` already activates the virtual environment, making an explicit call to `poetry shell` obsolete.


Makefile
```diff
@@ -83,11 +83,11 @@ poetry-shell:
 
 # Run evcc with python
 run-evcc:
-	$(shell which python) iso15118/evcc/main.py $(config)
+	poetry run python iso15118/evcc/main.py $(config)
```
tropxy, line 86: I had some issues in the past running it like that in a Mac.
Can someone confirm this? @mdwcrft @snorkman88

mdwcrft: This works for me

mdwcrft (approved):
LGTM

PR #315, opened 2023-10-25
Report session stop reason

shalinnijel2:
Report session stop reason. When [stop()](https://github.com/SwitchEV/iso15118/blob/e73c5903df92e419b8bc88a61e02626fa62721a9/iso15118/shared/comm_session.py#L376) is called, the reason is passed to evse_controller.


ikaratass (approved):
LGTM

PR #319, opened 2023-11-01
[JTT-291] - Relaxation of the PV limits

tropxy:
~As there are EVs that surpass the initial limitations imposed by DIN70121 and ISO 15118-2 to the Physical Values, it is important to provide a way to disable the validation, thus it was added a new config variable to enforce relaxation of the PV limits validations, called `IGNORE_PHYSICAL_VALUES_LIMITS` in settings.py~

EDIT: After an internal discussion, it was decided to remove the validations as in practice they are just a blocker for implementers. If later we come to the conclusion that some validations are necessary, those can be added case by case.

### DIFF
--- a/iso15118/shared/messages/datatypes.py
+++ b/iso15118/shared/messages/datatypes.py
@@ -11,6 +11,8 @@ from iso15118.shared.messages.enums import (
     UnitSymbol,
 )
 
+from iso15118.shared.settings import IGNORE_PHYSICAL_VALUE_LIMITS
+
 
 class PhysicalValue(BaseModel):
     """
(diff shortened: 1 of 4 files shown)


iso15118/shared/messages/datatypes.py
```diff
@@ -45,12 +48,15 @@ def validate_value_range(cls, values):
         value = values.get("value")
         multiplier = values.get("multiplier")
         calculated_value = value * 10**multiplier
-        if calculated_value > cls._max_limit or calculated_value < cls._min_limit:
+        if (
+            0 < cls._max_limit < calculated_value
+            or calculated_value < cls._min_limit < 0
+        ):
```
shalinnijel2, line 54: One question - was the intention here to limit calculated value from dropping below zero?
I'm not fully sure but I think the validator will never throw an error because 0 < cls._max_limit which will always be false and cls._min_limit < 0 which will also always be false?

shalinnijel2 (changes requested):

iso15118/shared/messages/datatypes.py
tropxy: the idea is, as long as the mix and min limits are 0, the validator wont ever throw an error, but, if we change the limits to something different than 0, then the validator will throw an error if the calculated value is outside of the boundaries.

shalinnijel2 (approved):

PR #335, opened 2023-12-06
JTT-458 Fix json logging in iso15118

shalinnijel2:
Broken in last release. Json logging was moved to be behind log EXI env variable.


ikaratass (approved):
lgtm

PR #342, opened 2023-12-08
Minor refactor exponent_value_conversion

shalinnijel2:
Minor change to share exponent value repr function between Physical Value and RationalNumber types.


ikaratass (approved):
LGTM

PR #358, opened 2024-01-08
jtt-550_ignore_incorrect_mo_root_on_disk

shalinnijel2:
Avoid failing if local chain validation fails in PaymentDetails. This will still be performed in the backend. Tests added.


santiagosalamandri (approved):
LGTM

PR #361, opened 2024-01-10
New EVCC config isAliveAfterSession introduced

touchlinux:
Whether or not the EVCC keeps running after a charging session is finished
Default 'True' should not affect EVCC behaviour
When set it to 'False', EVCC terminates by StopNotification on 'TCP connection closed'

Tested with modified `evcc_config_eim_dc.json` to verify EVCC is terminated gracefully.
```
{
	"supportedProtocols": [
		"ISO_15118_2"
	],
	"supportedEnergyServices": [
		"DC"
	],
	"energyTransferMode": "DC_extended",
	"isCertInstallNeeded": false,
	"useTls": false,
	"isAliveAfterSession": false
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,6 @@ Supported settings in EVCC configuration are given below:
 | sdpRetryCycles         | `1`                                                          | Indicates how often shall SDP (SECC Discovery Protocol) retries happen before reverting to using nominal duty cycle PWM-based charging |
 | maxContractCerts       | `3`                                                          | Maximum amount of contract certificates the EV stores.                                                                                 |
 | maxSupportingPoints    | `1024`                                                       | Indicates the maximum number of entries the EVCC supports within the sub-elements of a ScheduleTuple                                   |
-| isAliveAfterSession    | `True`                                                       | Whether or not the EVCC keeps running after a charging session is finished                                                             |
 
 
 The SECC and EVCC have been tested together under:
--- a/README.md
+++ b/README.md
@@ -173,6 +173,7 @@ Supported settings in EVCC configuration are given below:
 | sdpRetryCycles         | `1`                                                          | Indicates how often shall SDP (SECC Discovery Protocol) retries happen before reverting to using nominal duty cycle PWM-based charging |
 | maxContractCerts       | `3`                                                          | Maximum amount of contract certificates the EV stores.                                                                                 |
 | maxSupportingPoints    | `1024`                                                       | Indicates the maximum number of entries the EVCC supports within the sub-elements of a ScheduleTuple                                   |
+| isAliveAfterSession    | `True`                                                       | Whether or not the EVCC keeps running after a charging session is finished                                                             |
 
 
 The SECC and EVCC have been tested together under:
--- a/iso15118/evcc/comm_session_handler.py
+++ b/iso15118/evcc/comm_session_handler.py
@@ -306,7 +306,7 @@ class CommunicationSessionHandler:
 
         logger.info("Communication session handler started")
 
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -173,6 +173,7 @@ Supported settings in EVCC configuration are given below:
 | sdpRetryCycles         | `1`                                                          | Indicates how often shall SDP (SECC Discovery Protocol) retries happen before reverting to using nominal duty cycle PWM-based charging |
 | maxContractCerts       | `3`                                                          | Maximum amount of contract certificates the EV stores.                                                                                 |
 | maxSupportingPoints    | `1024`                                                       | Indicates the maximum number of entries the EVCC supports within the sub-elements of a ScheduleTuple                                   |
+| isAliveAfterSession    | `True`                                                       | Whether or not the EVCC keeps running after a charging session is finished                                                             |
```
shalinnijel2, line 176: Apologies I didn't clarify this under issues.  As the user can't proceed until the session is killed anyway, I don't see much value in making this a configurable field. The behaviour you have added to gracefully exit is the only change required. Do you have a use case in mind where keeping the session alive would be useful?

iso15118/evcc/comm_session_handler.py
```diff
@@ -550,7 +550,10 @@ async def get_from_rcv_queue(self, queue: asyncio.Queue):
                 elif isinstance(notification, StopNotification):
                     await cancel_task(self.comm_session[1])
                     del self.comm_session
-                    if not notification.successful:
+                    if notification.successful:
+                        if not self.config.is_alive_after_session:
+                            break
```
shalinnijel2, line 555: Following my suggestion from above, I think this would suffice. What do you think? 
```suggestion
                    if notification.successful:
                        break
```

shalinnijel2 (changes requested):

README.md
touchlinux: No, I don't have any use case in mind for keeping the session alive. I wanted to keep the session by default and stop the session when the configuration is used. However, it seems like the new configuration key is not necessary as stopping session can be the default behavior👍. I personally think, evcc should be terminated as a default after session, because evcc is not starting a new session again.

iso15118/evcc/comm_session_handler.py
```diff
@@ -306,7 +306,7 @@ async def __init__. Therefore, we need to create a separate async
 
         logger.info("Communication session handler started")
 
-        await wait_for_tasks(self.list_of_tasks)
+        await wait_for_tasks(self.list_of_tasks, asyncio.ALL_COMPLETED)
```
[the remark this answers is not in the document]
touchlinux: I think you are right, `asyncio.ALL_COMPLETED` is not necessary. `wait_for_tasks()` default argument `asyncio.FIRST_EXCEPTION` should suffice.
"FIRST_EXCEPTION | The function will return when any future finishes by raising an exception. If no future raises an exception then it is equivalent to ALL_COMPLETED." - https://docs.python.org/3/library/asyncio-task.html#waiting-primitives
Confirmed by testing without it.

iso15118/evcc/comm_session_handler.py
touchlinux: It looks good to me. Added another commit to fix all together.

shalinnijel2 (approved):

shalinnijel2:
Hi @touchlinux . All good. I'm merging the changes to master. Hope you don't mind. Thank you for your contribution. :).
I'll close the issue related to this as well. Feel free to reopen if there is anything more you would like to add to this.

PR #364, opened 2024-01-12
Added optional is_precharge flag to send_charging_command

shalinnijel2:
Added is_precharge flag to send_charging_command. Required as we use the same method during charging loop and precharge.


ikaratass (approved):
LGTM

PR #375, opened 2024-01-30
Fix for ChargeParameterDiscovery (TypeError occurred while processing message  in state ChargeParameterDiscovery : int() argument must be a string, a bytes-like object or a real number, not 'NoneType'. ).

ikaratass:
it is discovered in Testival Arnham that it is picking up the maximum voltage from DC even if it is a AC station. the error was: 
`2024-01-30 09:56:07,717  INFO	 - iso15118_service.evse_controller (1727): Session ended in ChargeParameterDiscovery (TypeError occurred while processing message  in state ChargeParameterDiscovery : int() argument must be a string, a bytes-like object or a real number, not 'NoneType'. ).`

### DIFF
--- a/iso15118/secc/controller/interface.py
+++ b/iso15118/secc/controller/interface.py
@@ -790,7 +790,9 @@ class EVSEControllerInterface(ABC):
             voltage_limit = session_limits.dc_limits.max_voltage
         else:
             ac_evse_charge_params = await self.get_ac_charge_params_v2()
-            voltage_limit = ac_evse_charge_params.evse_nominal_voltage.get_decimal_value()
+            voltage_limit = (
+                ac_evse_charge_params.evse_nominal_voltage.get_decimal_value()
+            )
         exponent, value = PhysicalValue.get_exponent_value_repr(voltage_limit)
         return PVEVSEMaxVoltageLimit(
             multiplier=exponent,
--- a/iso15118/secc/controller/interface.py
+++ b/iso15118/secc/controller/interface.py
@@ -786,13 +786,10 @@ class EVSEControllerInterface(ABC):
         - ISO 15118-2
         """
         session_limits = self.evse_data_context.session_limits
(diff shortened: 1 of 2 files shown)


OrangeTux (dismissed):

iso15118/secc/controller/interface.py
```diff
@@ -786,7 +786,13 @@ async def get_evse_max_voltage_limit(self) -> PVEVSEMaxVoltageLimit:
...
         """
         session_limits = self.evse_data_context.session_limits
-        voltage_limit = session_limits.dc_limits.max_voltage
+        if self.evse_data_context.current_type == CurrentType.DC:
+            voltage_limit = session_limits.dc_limits.max_voltage
+        else:
+            ac_evse_charge_params = await self.get_ac_charge_params_v2()
+            voltage_limit = (
+                ac_evse_charge_params.evse_nominal_voltage.get_decimal_value()
+            )
```
tropxy, line 795: this should work:

```python
async def get_evse_max_voltage_limit(self) -> PVEVSEMaxVoltageLimit:
        """
        Gets the max voltage that can be provided by the charger

        Relevant for:
        - ISO 15118-2
        """
        session_limits = self.evse_data_context.session_limits
        if self.evse_data_context.current_type == CurrentType.AC:
            voltage_limit = self.evse_data_context.nominal_voltage
        else:
            voltage_limit = session_limits.dc_limits.max_voltage
        exponent, value = PhysicalValue.get_exponent_value_repr(voltage_limit)
        return PVEVSEMaxVoltageLimit(
            multiplier=exponent,
            value=value,
            unit=UnitSymbol.VOLTAGE,
        )
```

ikaratass: Updated

ikaratass:
Changes moved to https://github.com/SwitchEV/iso15118/pull/376/files

PR #376, opened 2024-01-31
Fix for ChargeParameterDiscovery  get_evse_max_voltage

tropxy:
Fix for ChargeParameterDiscovery (TypeError occurred while processing message in state ChargeParameterDiscovery : int() argument must be a string, a bytes-like object or a real number, not 'NoneType'. )

Thank you @ikaratass for finding out where the bug was located and for the initial fix


ikaratass (approved):
LGTM

PR #387, opened 2024-03-13
Jtt 770 return ongoing if cs is not ready

shalinnijel2:
Return ongoing in ScheduleExchangeReq if evse is not ready.


ikaratass (approved):
LGTM

PR #388, opened 2024-03-18
Fixes for issue where the saved session context wasn't found on waking up.

shalinnijel2:
Changed the key to comm_sessions dict in secc comm sessions controller. The new key now uses the IP of the ev.


ikaratass (approved):
LGTM

PR #394, opened 2024-04-11
Added displayparams for ISO15118-20

heavyweight87:
Added DisplayParameters to DC/AC chargeloop req messages for ISO15118
It takes the SOC from the dc_ev_status


shalinnijel2:
Hey @heavyweight87 - I recently merged a PR that includes simulated soc and displayparameters from the EV. I used ev_ress_soc as well. It seems to me that includes what you wanted as well. If it does, then we could perhaps close this PR?

heavyweight87:
Great! Thanks

PR #395, opened 2024-04-12
Update .env.dev.local, ISO_15118_20_DC is not implemented for use

lwollinger:
I tried to use iso_20_DC but there was an error loading the protocol. The error was that ISO_20_DC was not implemented in .env.dev.docker


shalinnijel2 (approved):
Thank you for the fix! :)

PR #396, opened 2024-04-16
Moved contactor status check for dc to be after cable check

shalinnijel2:
Moved contactor status check for dc to be after a cable check response has been received.


ikaratass (approved):
LGTM

PR #398, opened 2024-04-22
fix typo in SECC interface.py

M4GNV5:
Just a small typo I found when planning the integration into a research project.

(force pushed to rebase change on top of current upstream master)


shalinnijel2 (approved):
Thank you for the fix! :)

PR #401, opened 2024-04-29
Share display params with CS.

shalinnijel2:
Share display params coming from EV with the CS. This doesn't apply to -2 AC though - as the request is empty.


iso15118/secc/controller/simulator.py
```diff
@@ -1090,3 +1090,9 @@ async def session_ended(self, current_state: str, reason: str):
         @param last_message: The last message that was either sent/received.
         """
         logger.info(f"Session ended in {current_state} ({reason}).")
+
+    async def send_display_params(self):
+        """
+        Share display params with CS.
+        """
+        pass
```
ikaratass, line 1098: would be good if we can add an info message

ikaratass (changes requested):

ikaratass (approved):
Approved.

PR #437, opened 2025-01-08
Relax stateC requirement in PowerDelivery(15118-20)

machinex85:
Relaxing the following requirement to a warning in this PR:
"[V2G20-847]: The EVCC shall signal CP State C or D no later than 250 ms after sending the first
PowerDeliveryReq with ChargeProgress equals "Start" within V2G communication session.."


ikaratass (approved):
Approved.

PR #460, opened 2026-04-05
Fix minor error in script comment (default password)

PaulSchulz:
Just noticed this as an internal inconsistency between comment and code.

### DIFF
--- a/iso15118/shared/pki/create_certs.sh
+++ b/iso15118/shared/pki/create_certs.sh
@@ -60,7 +60,7 @@ usage() {
     You need to provide the ISO 15118 version with the '-v' flag, choosing between the
     two admissible options stated above.
 
-    This script uses by default a password value of 12345, which can be modified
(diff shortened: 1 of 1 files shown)


iso15118/shared/pki/create_certs.sh
machinex85, line 141: The default password seems to be 12345. Am I missing something else?

machinex85 (changes requested):

iso15118/shared/pki/create_certs.sh
PaulSchulz: Sorry, my mistake.
One of the forks is using 123456 as their password and they hadn't changed the comment. I pushed to you, not realising my mistake.

PaulSchulz: FYI - Pull request for EVerest - https://github.com/EVerest/ext-switchev-iso15118/pull/61

